PR #72, opened 2026-04-14
Add identity audit logging

darshana-v:
## Summary

Identity mutations (create/update/delete) previously had no audit trail. Operators had no way to answer "who registered this agent?" or "who changed its trust level and when?".                           
                                                                                             
This PR adds an `identity_audit_logs` table that records every mutation with the acting user's ID (from the X-User-ID header, set by TenantContextMiddleware), a timestamp, and before/after JSONB snapshots of the identity row. 

---

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (does not change functionality, e.g. code style improvements, linting)
- [ ] This change requires a documentation update

---

## Testing
- [x] Manually tested
- [ ] Unit tests added/updated
- [ ] No tests required


darshana-v:
https://github.com/highflame-ai/highflame-admin/issues/1133

internal/service/identity.go
```diff
@@ -157,6 +167,13 @@ func (s *IdentityService) RegisterIdentity(ctx context.Context, req RegisterIden
 		Str("wimse_uri", identity.WIMSEURI).
 		Msg("Identity registered")
 
+	if s.auditRepo != nil {
+		newSnap, _ := json.Marshal(identity)
+		if err := s.auditRepo.Insert(ctx, req.AccountID, req.ProjectID, req.CallerUserID, identity.ID, auditActionCreate, auditStatusOK, nil, newSnap); err != nil {
```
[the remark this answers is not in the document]
darshana-v: Capturing failed mutations is out of scope for this initial PR - the goal here is to establish a baseline audit trail for successful identity lifecycle operations. Adding failure logging would require wrapping each mutation with an error-path audit write, which adds complexity and is better tracked as a separate follow-up issue.

migrations/008_identity_audit_logs.up.sql
```diff
@@ -0,0 +1,30 @@
...
+    status          VARCHAR(50)  NOT NULL DEFAULT 'SUCCESS',
+    old_data        JSONB,
+    new_data        JSONB,
+    created_at      TIMESTAMPTZ  NOT NULL DEFAULT NOW()
+);
+
+CREATE INDEX IF NOT EXISTS idx_identity_audit_logs_tenant
+    ON identity_audit_logs (account_id, project_id);
+
+CREATE INDEX IF NOT EXISTS idx_identity_audit_logs_identity
```
darshana-v: remove second and fourth indexes

abhisheksoamj:
The audit log should be generic and follow the same pattern we have in the admin service. 

Also, instead of passing callerUserID through every function signature, we should read it directly from the context using internalMiddleware.GetCallerName(ctx) at the point of writing the record in the database. 

Aren't we using triggers for writing to audit log table ?

internal/handler/audit.go
```diff
@@ -0,0 +1,101 @@
...
+	internalMiddleware "github.com/highflame-ai/zeroid/internal/middleware"
+	"github.com/highflame-ai/zeroid/internal/store/postgres"
+)
+
+// AuditLogResponse is the wire shape returned to Studio, matching the AuditLog
+// interface in packages/registry/src/types.ts.
+type AuditLogResponse struct {
+	AuditID     string                 `json:"audit_id"`
+	AccountID   string                 `json:"account_id"`
+	GatewayID   string                 `json:"gateway_id"`
```
abhisheksoamj: Are we still using GatewayID ? We can remove this from new designs

internal/handler/audit.go
```diff
@@ -0,0 +1,101 @@
...
+func (a *API) listAuditLogsOp(ctx context.Context, input *ListAuditLogsInput) (*ListAuditLogsOutput, error) {
+	tenant, err := internalMiddleware.GetTenant(ctx)
+	if err != nil {
+		return nil, huma.Error401Unauthorized("missing tenant context")
+	}
+
+	repo := postgres.NewAuditLogRepository(a.db)
+	entries, err := repo.List(ctx, tenant.AccountID, tenant.ProjectID, input.IdentityID)
+	if err != nil {
+		return nil, huma.Error500InternalServerError("failed to query audit logs")
```
abhisheksoamj: Wrong pattern, This means we are initialising repo every time we make a audit request. This will increase load on our DB and application. 

```
entries, err := repo.List(ctx, tenant.AccountID, tenant.ProjectID, input.IdentityID)
```

This should be called from service layer.

internal/handler/audit.go
```diff
@@ -0,0 +1,101 @@
...
+
+	out := &ListAuditLogsOutput{}
+	out.Body.Audits = make([]AuditLogResponse, 0, len(entries))
+	for _, e := range entries {
+		out.Body.Audits = append(out.Body.Audits, auditEntryToResponse(e))
+	}
+	return out, nil
+}
+
+func auditEntryToResponse(e postgres.AuditLogEntry) AuditLogResponse {
```
abhisheksoamj: This function should also be part of service

internal/handler/identity.go
```diff
@@ -170,7 +170,7 @@ func (a *API) createIdentityOp(ctx context.Context, input *CreateIdentityInput)
 		return nil, huma.Error400BadRequest("invalid sub_type for the given identity_type")
 	}
 
-	createdBy := internalMiddleware.GetCallerName(ctx)
+	callerUserID := internalMiddleware.GetCallerName(ctx)
```
abhisheksoamj: This should be called before we perform Insert/Update, similar to what we do in admin. 

Take a look at this example: 

internal/admin/module/application/schema.go
```
func (r *ApplicationSchema) BeforeAppendModel(ctx context.Context, query bun.Query) error {
	currentUserID, _ := model.GetUserIDFromContext(ctx)

	if _, ok := query.(*bun.InsertQuery); ok {
		r.CreatedBy = currentUserID
	}

	if _, ok := query.(*bun.UpdateQuery); ok {
		r.ModifiedBy = currentUserID
		r.ModifiedAt = time.Now()
	}

	return nil
}

```

internal/service/identity.go
```diff
@@ -74,7 +74,7 @@ type RegisterIdentityRequest struct {
 	Capabilities  json.RawMessage
 	Labels        json.RawMessage
 	Metadata      json.RawMessage
-	CreatedBy     string
+	CreatedBy string
```
abhisheksoamj: Please use the correct indentation for the struct object.

internal/service/identity.go
```diff
@@ -192,7 +192,7 @@ type UpdateIdentityRequest struct {
 	Capabilities  json.RawMessage
 	Labels        json.RawMessage
 	Metadata      json.RawMessage
-	Status        *domain.IdentityStatus
+	Status *domain.IdentityStatus
```
abhisheksoamj: Please use the correct indentation for the struct object.

internal/store/postgres/audit_read.go
abhisheksoamj, line 1: Why did we name this `store/postgres` ?

internal/handler/audit.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/danielgtaylor/huma/v2"
+
+	internalMiddleware "github.com/highflame-ai/zeroid/internal/middleware"
+	"github.com/highflame-ai/zeroid/internal/store/postgres"
+)
+
+// AuditLogResponse is the wire shape returned to Studio, matching the AuditLog
+// interface in packages/registry/src/types.ts.
+type AuditLogResponse struct {
```
abhisheksoamj: This struct should be declared in a different file (Please take a look at admin for reference )

internal/store/postgres/audit_read.go
```diff
@@ -0,0 +1,57 @@
...
+	AccountID    string          `bun:"account_id"     json:"account_id"`
+	ProjectID    string          `bun:"project_id"     json:"project_id"`
+	CallerUserID string          `bun:"caller_user_id" json:"caller_user_id"`
+	IdentityID   string          `bun:"identity_id"    json:"identity_id"`
+	Action       string          `bun:"action"         json:"action"`
+	Status       string          `bun:"status"         json:"status"`
+	OldData      json.RawMessage `bun:"old_data,type:jsonb"`
+	NewData      json.RawMessage `bun:"new_data,type:jsonb"`
+	CreatedAt    time.Time       `bun:"created_at"`
+}
```
abhisheksoamj, line 25: Where is the Table column in `AuditLogEntry` ? How will we capture audit info of rest of the tables ?

internal/store/postgres/audit_read.go
```diff
@@ -0,0 +1,57 @@
...
+
+// AuditLogRepository reads from identity_audit_logs.
+type AuditLogRepository struct {
+	db *bun.DB
+}
+
+// NewAuditLogRepository creates a new AuditLogRepository.
```
abhisheksoamj: These comment doesn't explain anything, it's not adding any value to the codebase

internal/store/postgres/audit_read.go
```diff
@@ -0,0 +1,57 @@
...
+		Model(&entries).
+		Where("account_id = ?", accountID).
+		OrderExpr("created_at DESC")
+
+	if projectID != "" {
+		q = q.Where("project_id = ?", projectID)
+	}
+	if identityID != "" {
+		q = q.Where("identity_id = ?", identityID)
+	}
```
abhisheksoamj, line 48: What about other filters like `Action`, `CallerUserId` ? 

What's `CallerUserId` ? Can't we just say `UserId` ?

internal/store/postgres/identity.go
```diff
@@ -132,7 +133,9 @@ func (r *IdentityRepository) List(ctx context.Context, accountID, projectID stri
 }
 
 // Update saves changes to an existing identity.
+// Sets modified_by from the request context so the audit trigger captures the acting user.
 func (r *IdentityRepository) Update(ctx context.Context, identity *domain.Identity) error {
+	identity.ModifiedBy = middleware.GetCallerName(ctx)
```
abhisheksoamj, line 137: We shouldn't do this here, I've added a comment please take a look at that. 
https://github.com/highflame-ai/zeroid/pull/72/changes#r3128932395

internal/store/postgres/identity.go
```diff
@@ -143,7 +146,15 @@ func (r *IdentityRepository) Update(ctx context.Context, identity *domain.Identi
 }
 
 // Delete removes an identity.
+// Sets modified_by before deleting so the AFTER DELETE trigger captures the acting user.
 func (r *IdentityRepository) Delete(ctx context.Context, id, accountID, projectID string) error {
+	if callerID := middleware.GetCallerName(ctx); callerID != "" {
```
abhisheksoamj, line 150: https://github.com/highflame-ai/zeroid/pull/72/changes#r3128932395

abhisheksoamj (changes requested):

internal/store/postgres/audit_read.go
darshana-v: :sob:

internal/store/postgres/audit_read.go
darshana-v: to differentiate between the one performing the operation and the one whose audit log has been made changes for (if it is stored for a same action). but since we have identity_id to identify the subject, will remove it

internal/store/postgres/identity.go
darshana-v: This does not point to any particular comment, do you mean getting modified by caller name from domain instead of middleware

abhisheksoamj: Oh okay, I just want to understand why are we using this function here ?

darshana-v: When a DELETE operation fires, the Postgres trigger only has access to OLD. This is a pre-stamp UPDATE to set OLD.modified_by to the current caller's ID immediately before the DELETE, so the trigger reads the correct actor for this operation. Without this, OLD.modified_by would reflect whoever last modified the identity, not who deleted it.

abhisheksoamj (approved):

PR #99, opened 2026-04-27
feat: propagate product claim from api_key grant to issued JWT

abhijitjavelin:
## Summary

API keys carry a `product` field in the database (`service_keys.product` — `ai_gateway`, `mcp_gateway`, etc.) but the api_key grant didn't forward it into the issued credential. Downstream services (Highflame Shield, Firehog) had no way to learn which product an inbound JWT was minted for and ended up hardcoding a default namespace, which caused per-project default-deny when a project's permits lived under a different namespace than the route happened to terminate at.

## Change

- Add `Product` field to `IssueRequest` (`internal/service/credential.go`).
- Set the `product` JWT claim alongside the existing `application_id` claim when non-empty (skipped on empty so legacy tokens are unchanged).
- `apiKeyGrant` (`internal/service/oauth.go`) forwards `sk.Product` into the issuance request.

## Why this matters end-to-end

Once this is released, firehog reads `claims.product` and routes the Shield request into the correct Cedar action namespace (`AIGateway::Action::*` for `ai_gateway` keys, `MCPGateway::Action::*` for `mcp_gateway` keys). That's the architectural fix that retires a set of project-wide hotfix permits we deployed today on dev1 + prod to work around the hardcoded namespace.

## Test plan

- [x] `go build ./...` — clean.
- [x] `go test ./internal/service/...` — no service-package tests yet (pre-existing).
- [ ] Post-release: bump zeroid in highflame-authn, mint an api_key JWT against a service key with `product = "mcp_gateway"`, decode and confirm the `product` claim is present.
- [ ] Then bump zeroid in highflame-firehog (after firehog [#105](https://github.com/highflame-ai/highflame-firehog/pull/105) merges) and verify Observatory events for MCP-keyed traffic show product = mcp_gateway.

### DIFF
--- a/internal/service/credential.go
+++ b/internal/service/credential.go
@@ -79,6 +79,11 @@ type IssueRequest struct {
 	UseRS256 bool
 	// ApplicationID is the optional application scope (set when API key is linked to an application).
 	ApplicationID string
+	// Product is the optional product scope (e.g. `ai_gateway`, `mcp_gateway`),
+	// set when the issuance source declares which product the credential is for.
+	// Downstream services (Shield, Firehog) read the `product` claim to pick the
+	// correct Cedar action namespace at evaluation time.
+	Product string
 	// SubjectOverride, when non-empty, replaces the default WIMSE URI as the JWT "sub" claim.
 	// Used for external principal exchange (sub = external user ID) and authorization_code
 	// (sub = authenticated user ID). For NHI grants, leave empty to use the WIMSE URI.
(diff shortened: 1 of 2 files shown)


abhijitjavelin:
Closing — going with a simpler approach: keep all firehog-routed traffic on the AIGateway namespace regardless of which product the service key was tagged with. The two real bugs were (1) policies app-scoped to specific apps but request had no app context, and (2) MCPContext.verified hardcoded false in firehog. Both are addressed by the data hotfix + firehog#107 — no need to plumb a product claim through the JWT.

PR #101, opened 2026-04-28
fix: serve SPIFFE trust bundle separately from JWKS (closes #43)

### LINKED ISSUE #43: JWT-SVID: JWKS use parameter is "sig" instead of "jwt-svid"
## Summary

The JWT-SVID specification requires that keys in the JWKS endpoint include \`"use": "jwt-svid"\` to clearly signal that these keys are intended for JWT-SVID verification. The current implementation sets \`"use": "sig"\` (the generic JWS value), which prevents conformant verifiers from distinguishing JWT-SVID keys from general signing keys.

## Location

JWKS key construction in the token issuance / key management layer (wherever the JWK is built and served at \`/.well-known/jwks.json\`).

## Impact

Conformant JWT-SVID verifiers MUST reject keys that do not carry \`"use": "jwt-svid"\`. This breaks interoperability with any standard SPIFFE workload API client.

## Fix

Change the \`use\` field in the published JWK from \`"sig"\` to \`"jwt-svid"\`.

## Reference

- [JWT-SVID spec §4](https://github.com/spiffe/spiffe/blob/main/standards/JWT-SVID.md#4-jwt-svid-bundle): "The \`use\` parameter … MUST be set to \`jwt-svid\`"

safayavatsal:
## Summary

- Publish `use=jwt-svid` on every JWK at `/.well-known/jwks.json` per [JWT-SVID §4](https://github.com/spiffe/spiffe/blob/main/standards/JWT-SVID.md#4-jwt-svid-bundle).
- Initial attempt flipped `use` in `addToKeySet` directly. That broke our own introspection path: `lestrrat-go/jwx`'s verifier (`jws/key_provider.go:113`) skips keys whose `use` is anything other than `"sig"`, so the in-memory keyset stopped verifying signatures.
- Split the concern: keep `use=sig` in the in-memory keyset (jwx-based verifiers, including our `/oauth2/token/introspect`, keep working) and rewrite `use` to `"jwt-svid"` at the JWKS handler boundary only.
- Normalize `use=jwt-svid → sig` on ingest in `pkg/authjwt`'s JWKS refresh so authjwt-based consumers keep working against the new bundle.
- Updates the regression assertion in `tests/integration/wellknown_test.go`.

Fixes #43

## Test plan

- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] `make test` passes (`TestJWKSEndpoint` asserts `use=jwt-svid`; `TestAuthjwtVerifiesZeroIDToken` and `TestAuthjwtAcceptsDefaultedAudience` confirm authjwt still works against the new bundle)
- [x] Manual: `curl http://localhost:8899/.well-known/jwks.json | jq '.keys[].use'` returns `"jwt-svid"` for both keys

### DIFF
--- a/internal/handler/wellknown.go
+++ b/internal/handler/wellknown.go
@@ -2,6 +2,8 @@ package handler
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"net/http"
 
 	"github.com/danielgtaylor/huma/v2"
@@ -10,8 +12,22 @@ import (
 
 // ── Well-known types ─────────────────────────────────────────────────────────
 
+// JWKSOutput is the RFC 7517 JSON Web Key Set published at
+// /.well-known/jwks.json. Keys carry use="sig" per RFC 7517 §4.2 so stock
+// OIDC / JWT libraries (PyJWT, jose, every WIF validator) accept the bundle.
+// SPIFFE-strict consumers should fetch /.well-known/spiffe-trust-bundle.json
+// instead.
 type JWKSOutput struct {
 	Body jwk.Set
+}
+
+// SPIFFETrustBundleOutput is the SPIFFE JWT-SVID trust bundle published at
+// /.well-known/spiffe-trust-bundle.json. The "keys" entries carry
+// use="JWT-SVID" per JWT-SVID §4. We use a generic map (not jwk.Set) because
+// we both rewrite "use" on each key and add the SPIFFE bundle envelope fields
+// ("spiffe_sequence", "spiffe_refresh_hint") that aren't part of RFC 7517.
+type SPIFFETrustBundleOutput struct {
+	Body map[string]any
 }
 
 type OAuthMetadataOutput struct {
@@ -25,9 +41,17 @@ func (a *API) registerWellKnownRoutes(api huma.API) {
 		OperationID: "jwks",
 		Method:      http.MethodGet,
 		Path:        "/.well-known/jwks.json",
-		Summary:     "JSON Web Key Set",
+		Summary:     "JSON Web Key Set (RFC 7517)",
 		Tags:        []string{"Discovery"},
 	}, a.jwksOp)
+
+	huma.Register(api, huma.Operation{
+		OperationID: "spiffe-trust-bundle",
+		Method:      http.MethodGet,
+		Path:        "/.well-known/spiffe-trust-bundle.json",
+		Summary:     "SPIFFE JWT-SVID trust bundle",
+		Tags:        []string{"Discovery"},
+	}, a.spiffeTrustBundleOp)
 
 	huma.Register(api, huma.Operation{
 		OperationID: "oauth-server-metadata",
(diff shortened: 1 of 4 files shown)


rsharath (approved):

safayavatsal:
@rsharath have fixed all the merge conflicts. can you please review.

rsharath:
| Vendor | Auth model | Status |
|---|---|---|
| **Anthropic Claude WIF** | `POST /v1/oauth/token` with `grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer`, SPIFFE listed provider | live today |
| **Azure Entra Workload Identity Federation** | Federated credential bound to external issuer JWKS | live today |
| **GCP Workload Identity Federation** | STS exchange → service-account access token | live today |
| **AWS IAM OIDC** | `AssumeRoleWithWebIdentity` | live today |

All four of these are JWT-SVID consumers in practice. The strict ones, and Azure is the canonical strict one today, with Anthropic likely to tighten as the SPIFFE-conformance bar lifts, reject JWKS entries whose `use` is anything other than `"JWT-SVID"` (or whose `use` field is missing). Without this PR, ZeroID's published JWKS at `/.well-known/jwks.json` carries `use="sig"` and is rejected by the strict consumers. **Tier 2b (the autonomous-agent + LLM provider story) cannot ship until this lands.**

### What this PR unlocks

1. **Anthropic Claude WIF interop.** Per-agent ZeroID assertions can be exchanged for `sk-ant-oat01-...` tokens with Anthropic-side service-account binding. End-to-end agent identity at the LLM provider — no shared tenant API key.
2. **Azure / GCP / AWS WIF interop.** Same shape. Workload federation against any cloud STS that accepts an external SPIFFE-compliant issuer.
3. **Closes the autonomous-agent gap (Firehog Tier 3 → Tier 2b).** Customers running CI-driven agents against WIF-capable SaaS no longer need the service-account-OAuth anti-pattern.
4. **Removes a blocker on the Firehog F5/F6 PRs.** Phase 4a delegation work for `auth_mode = "wif_exchange"` cannot reach an external vendor until the published JWKS passes their strict validators.

### Why the implementation is the right shape

The split — keep `use=sig` in-memory so jwx-based verifiers (including `/oauth2/token/introspect`) keep working, rewrite to `use=jwt-svid` at the JWKS handler boundary, normalize back to `sig` in `pkg/authjwt`'s ingest — is the surgical way to flip the published surface without touching every consumer's verifier path. It avoids the `WithKeySet` → `WithKey(alg, specificKey)` refactor that would otherwise be needed (since v4 of `lestrrat-go/jwx` relocated the same `use` filter to `jws/key_provider.go::keySetProvider.selectKey`).

rsharath:
## CI failure analysis + correction to my prior WIF-interop comment

### Root cause of the three CI failures

All three failing checks share one root cause: **stock JWT libraries enforce RFC 7517 §4.2's `use ∈ {\"sig\", \"enc\"}` allowlist and reject `use=\"jwt-svid\"`.** Specifically:

| Failing job | Library | Symptom |
|---|---|---|
| **Python SDK smoke** | PyJWT (`jwt.jwks_client.PyJWKClient`) | `get_signing_keys()` filters by `kty`/`use` and effectively drops keys with `use != \"sig\"` → `PyJWKClientError: The JWKS endpoint did not contain any signing keys` |
| **TypeScript SDK smoke** | `jose` (in `@highflame/sdk`) | `jose` validates JWK shape strictly: `Failed to import JWKS key: Invalid JWK \"use\" Parameter` |
| **highflame-regression-test** | PyJWT (same path) | Same Python SDK failure, re-run inside the regression suite |

The PR's split approach (keep `use=sig` in-memory, rewrite to `jwt-svid` at the handler, normalize back to `sig` in `pkg/authjwt` ingest) is correct for in-tree Go consumers, but it doesn't extend to external SDK consumers using stock JWT libraries. **Those libraries don't know to normalize \"jwt-svid\" → \"sig\"** — they just reject the key as malformed.

### Correction to my prior \"WIF interop\" comment

My earlier comment on this PR framed `use=jwt-svid` as a prerequisite for Anthropic / Azure / GCP / AWS WIF interop. **That framing is wrong and I need to retract it.**

WIF validators (Anthropic's `/v1/oauth/token`, Azure Entra federated credentials, GCP STS, AWS IAM OIDC) use stock OIDC / JWT libraries to fetch and validate the issuer's JWKS — **the exact same class of library that just failed in this PR's CI.** They want **RFC 7517-compliant** JWKS payloads with `use=\"sig\"` (or `use` omitted), not SPIFFE-strict `use=\"jwt-svid\"`. Publishing `use=jwt-svid` on `/.well-known/jwks.json` would likely *break* WIF interop, not enable it.

The conflict is real and structural:

| Consumer class | Wants |
|---|---|
| **SPIFFE-strict validators** (per JWT-SVID §4) | `use=\"JWT-SVID\"` |
| **Stock OIDC / JWT validators** (RFC 7517 §4.2) — includes Anthropic, Azure, GCP, AWS WIF | `use=\"sig\"` or `use` omitted |
| **Highflame's own SDKs today** (jose, PyJWT) | `use=\"sig\"` |

You can't satisfy both audiences with one endpoint and one payload shape.

### Suggested shape: dual endpoints, not in-place rewrite

The SPIFFE spec actually distinguishes a JWT-SVID **trust bundle** (SPIFFE-specific URL, SPIFFE-conformant payload) from a generic RFC 7517 JWKS. Recommend:

1. **Revert `/.well-known/jwks.json` to `use=\"sig\"`** — this is what every stock OIDC / JWT consumer wants, including WIF providers. Keep the rest of this PR's defensive plumbing (the ingest normalization in `pkg/authjwt` is harmless and good defense-in-depth).
2. **Add a new SPIFFE bundle endpoint** at e.g. `/spiffe/bundle.json` or `/.well-known/spiffe-trust-bundle.json` that serves the same keys with `use=\"jwt-svid\"`. Strict SPIFFE consumers point at this URL; stock JWT libs and WIF validators keep using `/.well-known/jwks.json`.
3. **Update `zeroid#43`'s acceptance criteria** to reflect this — the issue currently reads as if a single JWKS endpoint should satisfy both consumer classes, which the test failures show isn't achievable in practice.

This pattern is what real SPIFFE deployments use in production. Strict SPIFFE conformance via the bundle endpoint; back-compat with the OIDC / OAuth / WIF universe via the standard JWKS endpoint.

### TL;DR on the strategic framing

- **My earlier \"this unblocks WIF interop\" comment was wrong** — WIF validators are stock-JWT-library consumers and want `use=\"sig\"`. They're already satisfied by the pre-PR JWKS shape.
- **The SPIFFE-strict consumers that benefit from `use=\"jwt-svid\"`** are a separate audience (SPIRE-style validators, strict JWT-SVID verifiers in the SPIFFE ecosystem), not the cloud providers.
- **Two endpoints serves both cleanly.** One in-place rewrite breaks one or the other; the CI failures here are the proof.

Happy to author the dual-endpoint variant if useful. Sorry for the noise on the prior comment.

Issue #64, opened 2026-04-08
feat: Add CIBA (Client-Initiated Backchannel Authentication) support (RFC 9126)


rsharath:
## Summary

Add support for Client-Initiated Backchannel Authentication (CIBA) as defined in [RFC 9126](https://datatracker.ietf.org/doc/html/rfc9126) and the [OpenID Connect CIBA spec](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html). CIBA enables headless AI agents to initiate authentication on behalf of a user without requiring a browser redirect — the user approves via an out-of-band channel (push notification, email, or voice call).

## Motivation

CIBA is a critical flow for autonomous agent scenarios where:
- An agent needs user authorization but has no browser context (e.g., background orchestrators, CLI tools, MCP servers)
- A human-in-the-loop approval step is required before an agent acts on behalf of a user
- The agent operates in a headless environment (server-side, container, CI/CD pipeline)

This was identified as a gap compared to other agent identity platforms that support CIBA for headless agent approval flows.

### New Grant Type
- [ ] Register `urn:openid:params:grant-type:ciba` as a new grant type
- [ ] Implement backchannel authentication endpoint (`POST /oauth2/bc-authorize`)
- [ ] Implement polling mode: agent polls token endpoint with `auth_req_id`
- [ ] Implement ping mode: server notifies agent callback URL when user approves
- [ ] Support push mode (optional): server delivers token directly to agent callback

### Authentication Request
- [ ] Accept `login_hint` (user identifier — email, phone, user ID)
- [ ] Accept `scope` and `binding_message` (human-readable context for approval prompt)
- [ ] Accept `requested_expiry` for auth request TTL (default configurable, bounded by credential policy)
- [ ] Return `auth_req_id`, `expires_in`, and `interval` (polling interval)

### User Approval
- [ ] Define `BackchannelNotifier` interface for pluggable notification delivery:
  ```go
  type BackchannelNotifier interface {
      Notify(ctx context.Context, req BackchannelAuthRequest) error
  }
  ```
- [ ] Ship with no built-in notifier — users bring their own (push, email, SMS, Twilio, etc.)
- [ ] Expose `Server.SetBackchannelNotifier()` hook for registration
- [ ] Approval endpoint for user to accept/deny (`POST /oauth2/bc-authorize/{auth_req_id}/approve`)

### Token Issuance
[...]

PR #132, opened 2026-05-11
feat: CIBA ping mode (PR 2/3 for #64) — stacked on #131

safayavatsal:
## Summary

OpenID CIBA Core 1.0 §10.2 — **ping mode**. Second of three sequential dependent PRs that close issue #64.

> **⚠ This PR is stacked on #131 (PR 1).** It contains the PR 1 commit (`f00b692`) AND the PR 2 commit (`495f8c6`). **Do not merge this PR until #131 has merged.** Once #131 merges, the diff here will narrow to the PR 2 commit only — review will be straightforward at that point.
>
> **Merge order:** PR 1 (#131) → **PR 2 (this)** → PR 3.

Sibling discussion: https://github.com/highflame-ai/zeroid/issues/64#issuecomment-4418095786

## What changes vs PR 1

PR 1 shipped polling-mode CIBA end-to-end. PR 2 adds the **server-initiated callback** that lets clients skip the poll-loop:

1. Client registers with `client_notification_endpoint` (HTTPS) — the **per-client allowlist** for outbound POSTs.
2. Client posts `bc-authorize` with `client_notification_token` (the bearer the server will echo).
3. When the request transitions `pending → approved` (or `pending → denied`), the server POSTs `{auth_req_id}` to the registered endpoint with `Authorization: Bearer <client_notification_token>`.
4. Client then polls `/oauth2/token` and receives the access token (the token is NOT included in the ping — that's push mode, PR 3).

### Schema
- `migrations/015_ciba_ping_columns.{up,down}.sql`
  - `oauth_clients.client_notification_endpoint TEXT` — registered HTTPS allowlist
  - `backchannel_auth_requests.client_notification_token VARCHAR(1024)` — per-request bearer

### Domain / service
- `domain/token.go`: `OAuthClient.ClientNotificationEndpoint`
- `domain/backchannel_auth.go`: `BackchannelAuthRequest.ClientNotificationToken` (\`json:\"-\"\` — never serialised)
- `internal/service/oauth_client.go`: `validateNotificationEndpoint(...)` — HTTPS-only at registration
- `internal/service/backchannel.go`:
  - bc-authorize: switches to `ping` mode when `client_notification_token` is supplied AND the client has a registered HTTPS endpoint
  - `dispatchPing(ctx, row)` fires on Approve/Deny — bounded retry with exponential backoff + jitter, 4xx terminal, 5xx retried, records `last_notify_error`
  - `SetPingTransport(rt)`, `SetPingDispatchSync(bool)` — test-only

### Surface (Server)
- `Server.SetBackchannelPingTransport(http.RoundTripper)` — inject a capturing transport in tests
- `Server.SetBackchannelPingDispatchSync(bool)` — force synchronous dispatch in tests
- `OAuthClientConfig.ClientNotificationEndpoint` — pluggable via `EnsureClient`

### Tests (`tests/integration/ciba_ping_test.go`)
- **Happy path**: capturing RoundTripper asserts on URL, bearer, JSON body. Token does NOT contain `access_token` (push-mode guard).
- **Allowlist invariant**: `client_notification_token` without a registered `client_notification_endpoint` → 400 `invalid_request`.
- **HTTPS-only**: `http://` registration endpoint rejected at `EnsureClient`/`RegisterClient`.

## Security notes

- **Allowlist via registration**: per-request endpoints would defeat the allowlist; a leaked `client_notification_token` cannot redirect the outbound POST.
- **HTTPS at both layers**: validated at registration AND re-validated at request-time so a future bypass in one path doesn't silently degrade callbacks to plaintext.
- **Non-blocking dispatch**: `dispatchPing` runs in a goroutine with `context.WithTimeout` detached from the inbound request — caller cancellation can't kill an in-flight notification.
- **4xx terminal / 5xx retried**: client misconfiguration (bad bearer, wrong path) isn't retried; transient infrastructure faults are.
- **Token never persisted in JSON output**: `client_notification_token` carries \`json:\"-\"\` so list/get endpoints never echo it back.

## Issue #64 checklist delta vs PR 1

| Item | PR 1 | PR 2 |
|---|---|---|
| Ping mode | ⏳ | ✅ |
| Per-client URL allowlist | ⏳ | ✅ |
| Server retries on transient failure | ⏳ | ✅ |
| `client_notification_token` | ⏳ | ✅ |

## Test Plan

- [x] Build + vet clean (`go vet ./...`, `gofmt -l`)
- [x] New tests added — happy path, allowlist invariant, HTTPS-only registration
- [ ] CI runs the new tests (after #131 merges)
- [ ] Manual smoke after PR 1 + PR 2 land: register client with notification endpoint → bc-authorize with token → approve → observe ping → poll for access_token

### DIFF
--- a/domain/backchannel_auth.go
+++ b/domain/backchannel_auth.go
@@ -59,6 +59,7 @@ type BackchannelAuthRequest struct {
 	BindingMessage             string                      `bun:"binding_message,type:text"                    json:"binding_message,omitempty"`
 	NotificationMode           BackchannelNotificationMode `bun:"notification_mode,type:varchar(16)"           json:"notification_mode"`
 	ClientNotificationEndpoint string                      `bun:"client_notification_endpoint,type:text"       json:"client_notification_endpoint,omitempty"`
+	ClientNotificationToken    string                      `bun:"client_notification_token,type:varchar(1024)" json:"-"`
 	Status                     BackchannelStatus           `bun:"status,type:varchar(16)"                      json:"status"`
 	ApprovedSubjectID          string                      `bun:"approved_subject_id,type:varchar(255)"        json:"approved_subject_id,omitempty"`
 	ApprovedSubjectEmail       string                      `bun:"approved_subject_email,type:varchar(255)"     json:"approved_subject_email,omitempty"`
--- a/domain/token.go
+++ b/domain/token.go
@@ -85,6 +85,11 @@ type OAuthClient struct {
 	RedirectURIs []string `bun:"redirect_uris,array" json:"redirect_uris"`
 	Scopes       []string `bun:"scopes,array"       json:"scopes"`
 
+	// CIBA Core 1.0 — ping/push notification endpoint. Registered HTTPS URL
+	// the server POSTs to when a backchannel authentication request resolves.
+	// Empty when the client doesn't support ping mode (polling-only).
+	ClientNotificationEndpoint string `bun:"client_notification_endpoint" json:"client_notification_endpoint,omitempty"`
+
 	// Token lifetime (per-client, 0 = use server default)
 	AccessTokenTTL  int `bun:"access_token_ttl"  json:"access_token_ttl,omitempty"`
 	RefreshTokenTTL int `bun:"refresh_token_ttl" json:"refresh_token_ttl,omitempty"`
--- a/hooks.go
+++ b/hooks.go
@@ -60,6 +60,9 @@ type OAuthClientConfig struct {
 	SoftwareVersion         string
 	Contacts                []string
 	Metadata                json.RawMessage
+	// ClientNotificationEndpoint is the HTTPS callback CIBA ping mode posts to.
+	// Empty for clients that only use polling mode.
+	ClientNotificationEndpoint string
 }
 
 // TrustedServiceValidator checks whether the current request comes from a trusted
--- a/internal/handler/oauth.go
+++ b/internal/handler/oauth.go
@@ -285,13 +285,14 @@ func (a *API) revokeOp(ctx context.Context, input *OAuthRevokeInput) (*OAuthRevo
 // (clients in ZeroID are global to all tenants).
 type BcAuthorizeInput struct {
 	Body struct {
-		ClientID        string `json:"client_id" required:"true" doc:"OAuth client initiating the backchannel request"`
-		AccountID       string `json:"account_id" required:"true" doc:"Tenant account ID"`
-		ProjectID       string `json:"project_id" required:"true" doc:"Tenant project ID"`
-		LoginHint       string `json:"login_hint" required:"true" doc:"User identifier (email, phone, user_id)"`
-		Scope           string `json:"scope,omitempty" doc:"Requested scopes (space-delimited)"`
-		BindingMessage  string `json:"binding_message,omitempty" doc:"Human-readable context shown to the user during approval"`
-		RequestedExpiry int    `json:"requested_expiry,omitempty" doc:"Auth-request TTL in seconds; bounded by server default"`
+		ClientID                string `json:"client_id" required:"true" doc:"OAuth client initiating the backchannel request"`
+		AccountID               string `json:"account_id" required:"true" doc:"Tenant account ID"`
+		ProjectID               string `json:"project_id" required:"true" doc:"Tenant project ID"`
+		LoginHint               string `json:"login_hint" required:"true" doc:"User identifier (email, phone, user_id)"`
+		Scope                   string `json:"scope,omitempty" doc:"Requested scopes (space-delimited)"`
+		BindingMessage          string `json:"binding_message,omitempty" doc:"Human-readable context shown to the user during approval"`
+		RequestedExpiry         int    `json:"requested_expiry,omitempty" doc:"Auth-request TTL in seconds; bounded by server default"`
+		ClientNotificationToken string `json:"client_notification_token,omitempty" doc:"Bearer the server echoes in the ping callback (required for ping mode)"`
 	}
 }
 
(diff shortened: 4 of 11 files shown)


safayavatsal:
**Changes since last commit:**

Addressed the Gemini review comments, merged the PR 1 branch (which carries the PR 1 review fixes plus the upstream-main merge), and resolved migration-number collisions.

### Direct fixes (this PR's own surface)

| Comment | Status | Notes |
|---|---|---|
| `r3216971175` — backoff timeout under-allocated | ✅ Fixed | Detached context now budgets `PingTimeout*(maxRetries+1) + baseDelay*(2^maxRetries - 1)` (the actual geometric sum of exponential retry waits) instead of the previous linear approximation, so the final attempt cannot be cut off mid-flight. |
| `r3216971184` — `time.After` leaks + jitter unnecessary | ✅ Fixed | Replaced `time.After` with `time.NewTimer` + explicit `Stop()` so timers are collectable as soon as the select returns. Dropped jitter entirely — this is a single-process retry loop against a private dependency, so the thundering-herd scenario jitter targets doesn't apply, and the non-determinism only hurts test reproducibility. `jitterMillis` helper removed. |

### Superseded by PR 1 fixes (now merged into this branch)

The PR 1 branch already shipped stricter answers to two of the comments here:

| Comment | What PR 1 did instead |
|---|---|
| `r3216971146` + `r3216971164` — UTF-8-safe truncation of `binding_message` | PR 1 stopped truncating altogether: oversized `binding_message` now returns a wrapped `ErrInvalidBindingMessage` sentinel (handler maps via `errors.Is` to 400). Strictly better than the safe-truncate suggestion — no codepoint splits to worry about because no string is ever sliced. |
| `r3216971137` + `r3216971188` — wrap `IssueCredential`+`MarkIssued` in a tx via `bun.IDB` refactor | PR 1's `Redeem` flips `approved → issued` BEFORE minting. The conditional UPDATE in `MarkIssued` (`WHERE status='approved'`) is the at-most-once invariant: only one polling thread reaches `IssueCredential`; a concurrent caller gets `affected=0` and `access_denied`. This achieves single-use enforcement without the cross-repo `bun.IDB` refactor. Tx-based approach can still land later as a defence-in-depth improvement, but the race itself is closed today. |

### Plus: bonus from the PR 1 merge

- `dispatchPing` now also parents on the service `svcCtx` (introduced in PR 1's lifecycle-context fix), so `Server.Shutdown` cancels in-flight pings instead of letting them outlive the listener close.

### Migration renumbering

Upstream `main` landed `014_attestation_policies` and `016_identity_risk_metadata` while these PRs were open. To resolve the prefix collision:

- `014_backchannel_auth_requests` → **`015_backchannel_auth_requests`** (PR 1)
- `015_ciba_ping_columns` → **`017_ciba_ping_columns`** (this PR)

golang-migrate requires unique numeric prefixes; non-contiguous numbering is supported. Renames preserve content; only the filename prefix changed.

`go vet ./...` and `gofmt -l .` clean post-merge.

safayavatsal:
@rsharath resolved all the merge conflicts

rsharath:
Hey @safayavatsal — rebased your branch onto the new `main` (now that #131 is merged) and force-pushed since the migration-number collision was blocking. Your commit is preserved with full authorship on the single rebased commit. Here's what changed:

## Rebase summary

**Dropped** (redundant after #131 merge):
- `f00b692` — old polling-mode commit (squashed into main as #131)
- `fe5d8fb` — Gemini review fixup (its three fixes — race condition, binding_message truncation sentinel, svcCtx lifecycle — all landed in `main` via #131's merge commit; cherry-picking it produced no net diff)
- 5 merge-from-main commits (noise from periodic merges into the branch)

**Kept** (your actual ping-mode work):
- `495f8c6` → rebased as `30119b8` → now `feat: CIBA ping mode (PR 2/3 for #64)`

**Renamed**: migration `015_ciba_ping_columns` → `020_ciba_ping_columns` to slot in after `019_backchannel_auth_requests` (now occupied by the merged polling work).

## Three small fixes applied during the rebase

I caught these during a pre-push review — happy to revert any if you'd rather make the call yourself:

1. **`dispatchPing` now parents on `svcCtx`** (was using `context.Background()`).
   - The `dispatchNotifier` path in #131 was correctly updated to parent on `svcCtx` so `Server.Shutdown` → `Stop()` can cancel in-flight retries. `dispatchPing` was written before that pattern landed and didn't get the same treatment. Mirrored the `dispatchNotifier` pattern: read `parent := s.svcCtx` under RLock, drop the dispatch if `parent == nil` (Stop already called).

2. **Drain HTTP response body before `Close()`**: added `io.Copy(io.Discard, resp.Body)` before `resp.Body.Close()`. The original `func() { defer resp.Body.Close() }()` closes without draining, which hurts HTTP/1.1 keepalive connection reuse on retries. Minor perf improvement.

3. **Migration file headers** updated from `015_ciba_ping_columns` → `020_ciba_ping_columns` so the SQL comment matches the filename. Also updated the commit-message body's reference to migration `020`.

## What's still missing (future PRs, not blocking this one)

- **No SSRF guard on the HTTPS endpoint**: `validateNotificationEndpoint` enforces HTTPS but doesn't block private IP ranges (`https://10.0.0.5/`, `https://localhost/`, etc.). For v1 this is acceptable because the endpoint is admin-registered (trusted operator-side), but worth filing a hardening follow-up for multi-tenant deployments where untrusted operators could register endpoints.

## Verification

- Build: `GOEXPERIMENT=jsonv2 go build ./...` ✓ clean
- Vet: `GOEXPERIMENT=jsonv2 go vet ./...` ✓ clean
- Authorship: `git log -1` shows you as Author, me as Committer
- CI will re-run automatically on the force-push

Thanks for the contribution — the design here (per-client allowlist, two-tier HTTPS check, exponential backoff with jitter, 4xx-terminal/5xx-retry, last_notify_error recording, capturing RoundTripper in tests) is solid. The bug I caught was a missed wiring detail, not a design issue.

Once CI is green I'll add an approval review so the branch protection unblocks. Then we can do #133 (push mode) with the same flow — should be smoother now that the foundation + ping mode are in.

rsharath (approved):
Approved. Reviewed the rebased diff pre-push (see #issuecomment-4444241290 above for the three fixes applied during rebase). Build + vet clean, CI 10/10 green.Squash-merging.

PR #176, opened 2026-06-02
feat: add RFC-style spec for ZeroID OAuth/OIDC extensions

rsharath:
## What

Adds a consolidated Internet-Draft-style normative specification at `docs/spec/zeroid-oauth-extensions.md` documenting **every extension ZeroID layers on top of the baseline specs** it implements (OAuth 2.1 / OIDC / RFC 8693 / RFC 9449 / OpenID CIBA Core 1.0 / RFC 9396 / RFC 7591-7592 / RFC 8414 / RFC 9728). Baseline conformance is explicitly out of scope, this covers the deltas only, at the wire level, so an independent implementer or resource server can interoperate without reading the source.

## Contents (14 sections)

- **Identity model & WIMSE URIs** — `spiffe://{domain}/{account}/{project}/{type}/{id}`, 2048-byte cap, SPIFFE-safe charset, and the `identity_type` / `sub_type` / `trust_level` / `status` vocabularies.
- **JWT claim extensions** — exact name/type/presence for `account_id`, `project_id`, `grant_type`, `mission_id`, the identity claims, `owner_user_id`, `delegation_depth`, `act`, `cnf.jkt`, `token_exchange`/`trusted_by`, gated `role`/`privilege_scope`, and the CIBA claims.
- **RFC 8693** — three-way scope attenuation, depth capping, actor PoP requirement, `mission_id` propagation, external-principal exchange, cascade revocation.
- **CIBA** — `group_hint` extension param + RAR on bc-authorize + delivery modes.
- **DPoP** — the `bh` body-hash claim, algorithm allow-list, validation order.
- **Reserved-claims gating**, **CAE/SSF signals**, **non-standard endpoints** (delegation explorer / signals / WIMSE proof tokens), **discovery metadata extensions**, **security considerations**, **IANA-style registry**, **references**.

## Notes

- Docs-only change (plus a one-line README pointer in the Standards section for discoverability). No code, no migrations, no API changes.
- **Every wire-level fact was verified against the source.** That caught one real discrepancy worth flagging for review: the CIBA `grant_type` claim serializes as the full URN `urn:openid:params:grant-type:ciba`, unlike the other six grants which use short forms — the spec documents this asymmetry rather than hiding it.
- Cross-repo closing keywords link but do not auto-close; the authn issue will need a manual close once merged.


docs/spec/zeroid-oauth-extensions.md
```diff
@@ -0,0 +1,730 @@
...
+claim only when &gt; 0 (Section 4.3).
+
+### 5.3 NHI delegation requires an actor proof-of-possession
+
+For NHI → NHI delegation, both `subject_token` (the orchestrator's active
+credential) and `actor_token` (a JWT assertion the sub-agent self-signs with the
+private key matching its registered public key) are **REQUIRED**. ZeroID
+verifies the actor assertion's `iss`/`sub` equal the actor identity's WIMSE URI
+before issuing. The issued token's `act.sub` is set to the orchestrator's WIMSE
+URI (Section 4.4 case 1).
```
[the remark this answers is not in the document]
rsharath: Applied in 5b32284 — good catch. This matches what the implementation already enforces: `oauth.go:499-509` explicitly requires `sub` and `exp` on the `actor_token` (citing RFC 7523 §3 / RFC 8693 §1.2, since jwx honors but does not require them). §5.3 now documents the full actor_token contract: asymmetric alg, `iss` = actor WIMSE URI, `aud` = issuer, and required `sub`/`exp`.

docs/spec/zeroid-oauth-extensions.md
```diff
@@ -0,0 +1,730 @@
...
+- `external_id` — the registrant-supplied stable identifier for the principal,
+  unique within `(account_id, project_id, identity_type)`.
+
+Constraints:
+
+- Each path segment **MUST** consist only of the characters `a–z A–Z 0–9 . - _`
+  (SPIFFE §2.3 path-segment safe set). Registration **MUST** reject any segment
+  outside this set.
+- The assembled URI **MUST NOT** exceed **2048 bytes**. Registration **MUST**
+  reject a longer URI rather than persist a non-conformant subject.
```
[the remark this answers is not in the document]
rsharath: Partially applied. The suggestion to forbid uppercase `A-Z` would make the spec contradict the implementation: ZeroID's `ValidateSPIFFEPathSegment` (`domain/identity.go:474-489`) explicitly permits `a-z A-Z 0-9 . - _`, so uppercase is allowed by design — this spec documents actual behavior, so I kept `A-Z`. I did adopt the valid part of your point: §3.1 now notes that the segment safe-set excludes `/ ? # @ :`, so empty segments, query strings, fragments, user-info, and embedded ports cannot appear in a segment. The `trust_domain` is server configuration (not request input) and is not re-validated per request, so I did not add a per-request "no port in trust domain" MUST that the code does not enforce.

docs/spec/zeroid-oauth-extensions.md
```diff
@@ -0,0 +1,906 @@
...
+## 1. Introduction
+
+OAuth 2.x and OIDC authenticate a *human to a service*. They have no native
+model for a credential whose holder is an autonomous agent, for one agent
+delegating an attenuated subset of its authority to another, for the resulting
+multi-hop chain to remain cryptographically attributable, or for an agent to
+pause mid-task and obtain a human's out-of-band approval.
+
+ZeroID exists to solve **Agent Identity**. Human identity (the domain of
+OAuth/OIDC IdPs) and general workload / machine identity (the domain of
```
adeinega: Strictly speaking, OAuth2 is not (solely) concerned with human identity, as an example, the JWT Bearer and Client Credentials grant types barely have anything to do with it.

I would like to add that the OpenID Connect Foundation generally tends to use “OpenID Connect Providers” or OPs, rather than OIDC IdPs.

rsharath: thanks @adeinega for the comment. 

The intro now frames the gap more precisely. OAuth's non-human story models a flat machine credential, and what's missing is the agent-shaped layer on top (delegation chains, attenuation, cross-hop attribution, out-of-band approval).
  
Also switched "OIDC IdPs" -> "OpenID Providers" per OIDF usage (kept "enterprise IdPs" alongside, since the sentence is about the human-identity ecosystem broadly).

adeinega: @rsharath, 2 more cents from me... the same applies to abbreviations "OIDC" and "OIDC token", OpenID Connect Foundation refers to them as "OpenID Connect" and "OpenID Connect ID Token" in their specs such as [OpenID Connect Core 1.0 Core](https://openid.net/specs/openid-connect-core-1_0.html), and also their official communications say [Identity Management for Agentic AI](https://openid.net/wp-content/uploads/2025/10/Identity-Management-for-Agentic-AI.pdf). IETF OAuth specs do the same, for example, https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-15.html.

Now, the question is, does it even matter? I would say “yes” assuming your goal for ZeroID to adhere to these standards and use the same language and terms.

rsharath: Fair point @adeinega. Applied in https://github.com/highflame-ai/zeroid/commit/669b34fe9b8256d4bbbb4f8053b27b31b4f122c6.

Scoped it to the conceptual references where the doc cites the standard itself: the title, the Sec1 intro, and the "builds on … OpenID Connect, RFC 8693" primitives list now read OpenID Connect per OIDF/IETF spec usage.

I deliberately left Section 10 (WIF) as OIDC token. Those are upstream federation tokens like GitHub Actions, GCP Workload Identity, k8s projected SA tokens Likewise the code identifiers (oidc_token proof type, OIDCPolicyConfig, OIDCIssuerConfig) stay verbatim since the spec documents the implementation's actual field names.

rsharath:
Synced the spec with behavior merged to `main` since the last staleness pass (`571455f`):

- **§5.6 (new): direct OIDC ID-token federation** — the `subject_token_type=id_token` exchange merged in #211 (verifies an upstream OP's ID token against a configured JWKS allowlist; required fail-closed `allowed_accounts` tenant binding; alg pinning; `max_token_age`; SSRF-guarded JWKS fetch). Cascade revocation renumbered §5.6 → §5.7.
- **§4.6**: `user_id_iss` provenance claim + `auth_time`/`acr`/`amr` passthrough semantics.
- **§8**: added `user_id_iss` to the reserved-claims blocklist (the spec was factually out of step with the code's `reservedClaims` map).
- **§3**: identity status machine now includes `expired` (#208).
- **§10 / §12 / §13**: disambiguated workload vs user OIDC federation, added a direct-federation security consideration, and registered the new claims + `id_token` subject-token-type.

No discovery-metadata change — #211 added no new `.well-known` fields; the `id_token` mode rides the already-advertised `token_exchange` grant.

jalbrethsen-highflame (approved):

PR #177, opened 2026-06-03
feat: add /oauth2/authorize endpoint + PrincipalResolver hook

saucam:
## Summary

Adds the upstream half of the OAuth 2.0 + RFC 7636 PKCE `authorization_code` grant — `POST /oauth2/authorize` — that mints the auth-code JWT zeroid already consumes at `/oauth2/token`. Closes the asymmetry where zeroid owned `decodeAuthCodeJWT` + `authCodeRepo.Consume` but never shipped the issuance side.

Together with the new `PrincipalResolver` hook (peer of `ClaimsEnricher`, `GrantHandler`, `BackchannelNotifier`), deployers can plug zeroid into any principal-authentication source (`api_key` today, browser session / mTLS / SPIFFE chain tomorrow) without forking the OAuth contract.

## New public surface (zeroid top-level)

```go
// Hooks
type Principal struct {
    AccountID, ProjectID, UserID, OrgID string
    Scopes []string
}

type AuthorizeRequest struct {
    ClientID, RedirectURI, ResponseType,
    CodeChallenge, CodeChallengeMethod,
    State, Scope string
    Form, Header, Cookie func(name string) string
}

type PrincipalResolver func(ctx context.Context, req *AuthorizeRequest) (*Principal, error)
var  ErrPrincipalNotApplicable = errors.New(...)

// Server
func (s *Server) RegisterPrincipalResolver(name string, r PrincipalResolver)
```

## New public surface (service package)

```go
// Authorization-code issuance (the missing half)
type IssueAuthCodeRequest struct {
    ClientID, RedirectURI, CodeChallenge, CodeChallengeMethod string
    AccountID, ProjectID, UserID, OrgID string
    Scopes []string
}
func (s *OAuthService) IssueAuthCode(ctx context.Context, req IssueAuthCodeRequest) (string, error)

// API-key resolution (extracted from apiKeyGrant via refactor — no duplicate logic)
type APIKeyResolution struct {
    AccountID, ProjectID, UserID, KeyID string
    Scopes []string
}
func (s *OAuthService) ResolveAPIKey(ctx context.Context, apiKey string) (*APIKeyResolution, error)
```

## Design decisions worth flagging

- **`AuthorizeRequest` is a typed snapshot, not `*http.Request`.** Resolvers see typed `Form`/`Header`/`Cookie` accessors; never net/http. Keeps the hook consistent with zeroid's existing typed-input pattern (`TokenRequest`, `BcAuthorizeInput`) instead of letting one hook leak transport details.
- **`ResolveAPIKey` returns a narrow `APIKeyResolution`, not `*domain.Identity`.** The full Identity stays internal (`apiKeyContext`); public callers get a stable projection that doesn't break if the Identity struct changes.
- **`/oauth2/authorize` is a chi route, not Huma.** The endpoint is 302-on-success with a resolver-dependent principal-credential field set — neither shape fits a static JSON I/O schema cleanly. Mirrors `/oauth2/token/verify` which is also chi-not-Huma.
- **`Principal`, `AuthorizeRequest`, `PrincipalResolver` are defined in `internal/service` and aliased at zeroid top level.** Necessary because `internal/handler` can't import the top-level `zeroid` package (circular). Aliases (`type Principal = service.Principal`) make the public API identical.
- **`/oauth2/authorize` is intentionally NOT in `OAuthFormEndpoints`.** That middleware rewrites form→JSON for Huma handlers; my chi handler reads the form-encoded body directly.

## Refactor: `apiKeyGrant` → `resolveAPIKeyContext` + grant-specific work

`apiKeyGrant` was 125 lines doing both api_key resolution AND token minting. Extracted the resolution piece (~50 lines) into `resolveAPIKeyContext`, which both `apiKeyGrant` and the new public `ResolveAPIKey` call. No behaviour change for `apiKeyGrant` — same lookup, same identity-state checks, same error shapes.

## Validation gates in `IssueAuthCode`

Each enforced before any DB hit / JWT mint:

1. Required fields: `client_id`, `redirect_uri`, `code_challenge`, `code_challenge_method`, `account_id`
2. `code_challenge_method` MUST be `S256` (plain rejected — RFC 7636 §4.2 deprecation + OAuth 2.1 removal)
3. HMAC secret must be configured (defensive 500 if not)
4. Client exists + is active (`oauthClientSvc.GetPublicClient`)
5. Client has `authorization_code` in its registered `grant_types`
6. `redirect_uri` matches one of the client's registered URIs (`normalizeLoopback` for RFC 8252 §7.3 equivalence)
7. Scope intersection: requested ∩ client's registered scopes

Same checks `authorizationCode` (the consume side) runs at exchange time — applied here too so a client without the grant cannot even obtain a code, not just fail at exchange.

## Tests

**Unit** — `internal/service/authcode_issue_test.go` (5 tests, pure, no DB):
- `TestMintAuthCodeJWT_RoundTripsThroughDecoder` — load-bearing contract pin
- `TestMintAuthCodeJWT_OptionalClaimsOmittedWhenEmpty`
- `TestMintAuthCodeJWT_RejectsMissingRequiredFields` (table-driven)
- `TestMintAuthCodeJWT_RespectsCustomJTI`
- `TestMintAuthCodeJWT_DefaultExpHonoursAuthCodeTTL`
- `TestRedirectURIAllowed` (table-driven)

**Integration** — `tests/integration/oauth_authorize_test.go` (11 tests, real Postgres via testcontainers):
- `TestAuthorize_HappyPath` — full mint → exchange → introspect round-trip
- `TestAuthorize_StateOmitted` — state= absence vs empty string
- `TestAuthorize_MissingRequiredFields` (table-driven on every required form field)
- `TestAuthorize_NonS256ChallengeMethod`
- `TestAuthorize_NonCodeResponseType`
- `TestAuthorize_UnknownClient`
- `TestAuthorize_RedirectURIMismatch`
- `TestAuthorize_RedirectURILoopbackEquivalence`
- `TestAuthorize_NoPrincipalMatched` (chain-end with `ErrPrincipalNotApplicable`)
- `TestAuthorize_ResolverError` (non-sentinel error path)
- `TestAuthorize_IssuedCodeShapeMatchesDecoder` — explicit JWT shape pin
- `TestAuthorize_PKCERoundTrip` — extra verifier→challenge round-trip

A stub `PrincipalResolver` is registered in `TestMain` reading magic `test_principal_*` form fields so a single resolver covers every test shape without per-test registration.

## Migration notes for consumers

`highflame-authn` is the first consumer. Today its `highflame-authn#93` PR (currently in draft) carries its own `IssueAuthCode` + `/oauth2/authorize` handler — that will be replaced by:

- A ~30-line `api_key` `PrincipalResolver` in authn that calls `zeroid.ResolveAPIKey` and projects to `*zeroid.Principal`
- A one-line `srv.RegisterPrincipalResolver("api_key", resolver)` in authn's `cmd/server/main.go`
- Deletion of the duplicate `IssueAuthCode` + ugly `httptest.NewRecorder` roundtrip currently in authn

The companion regression test (`highflame-regression#85`) doesn't change — same wire contract.

## Test plan

- [ ] CI passes (golangci-lint + unit + integration with testcontainers)
- [ ] Spot-check OpenAPI spec — `/oauth2/authorize` is intentionally absent from it (chi route); document this in CHANGELOG if reviewers want it
- [ ] Review the `PrincipalResolver` abstraction — does it cover the cases zeroid wants to support for v2 (browser session, mTLS, SPIFFE chain)?
- [ ] Review the refactor of `apiKeyGrant` — no behaviour change intended; the test suite should catch any regression

### DIFF
--- a/hooks.go
+++ b/hooks.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/highflame-ai/zeroid/domain"
+	"github.com/highflame-ai/zeroid/internal/service"
 )
 
 // ClaimsEnricher is called during JWT issuance to add custom claims.
@@ -38,6 +39,37 @@ type GrantRequest struct {
 	Role           string
 	PrivilegeScope []string
 }
+
+// Principal is the resolved caller at /oauth2/authorize — the tenant +
+// user binding that gets baked into the issued authorization code JWT.
+// Re-exported from internal/service so deployer code stays at the
+// top-level zeroid public surface; both names refer to the same type.
+//
+// See internal/service/principal.go for the canonical doc comment.
+type Principal = service.Principal
+
+// AuthorizeRequest is the typed, read-only snapshot of the parsed
+// /oauth2/authorize request handed to every PrincipalResolver. Resolvers
+// see this — never net/http types — so the extensibility hook stays
+// consistent with zeroid's other typed-struct boundaries.
+//
+// See internal/service/principal.go for the canonical doc comment.
+type AuthorizeRequest = service.AuthorizeRequest
+
+// PrincipalResolver authenticates the caller at /oauth2/authorize.
+// Registered via Server.RegisterPrincipalResolver and tried in
+// registration order; the first resolver to return a non-nil Principal
+// wins. Return ErrPrincipalNotApplicable to defer to the next resolver;
+// any other error fails the request with 401 invalid_client.
+//
+// See internal/service/principal.go for the canonical doc comment.
+type PrincipalResolver = service.PrincipalResolver
+
+// ErrPrincipalNotApplicable is the sentinel returned by a
+// PrincipalResolver that does not apply to the current request. zeroid
+// moves to the next registered resolver; when every resolver returns
+// this sentinel, the request fails with 401 invalid_client.
+var ErrPrincipalNotApplicable = service.ErrPrincipalNotApplicable
 
 // AdminAuthMiddleware is an optional middleware applied to the admin API router.
 // When set, every request to the admin port passes through this middleware before
--- a/hooks.go
+++ b/hooks.go
@@ -71,6 +71,17 @@ type PrincipalResolver = service.PrincipalResolver
 // this sentinel, the request fails with 401 invalid_client.
 var ErrPrincipalNotApplicable = service.ErrPrincipalNotApplicable
 
+// ErrNoResolversRegistered is the sentinel surfaced by zeroid when
+// /oauth2/authorize is reached but no PrincipalResolver has been
+// registered via Server.RegisterPrincipalResolver. The handler maps
+// this to 503 Service Unavailable so the deployer sees a clear
+// "you forgot to wire this up" signal rather than an ambiguous 401.
+//
+// Deployers don't typically observe this sentinel directly — it's
+// emitted by zeroid's chain walker and consumed by the handler. The
+// re-export exists so deployer tests can match on it via errors.Is.
+var ErrNoResolversRegistered = service.ErrNoResolversRegistered
+
 // AdminAuthMiddleware is an optional middleware applied to the admin API router.
 // When set, every request to the admin port passes through this middleware before
 // reaching any handler. Use this to add authentication (Bearer JWT, mTLS, API key,
--- a/internal/handler/authorize.go
+++ b/internal/handler/authorize.go
@@ -1,0 +1,276 @@
+package handler
+
+// /oauth2/authorize — the upstream half of the OAuth 2.0 + PKCE
+// authorization_code grant. Mounted as a plain chi route (not Huma)
+// because:
+//
+//  1. The principal-credential field set is resolver-dependent —
+//     api_key today, session cookie / mTLS tomorrow — and doesn't fit
+//     a static OpenAPI input schema. Huma's value (declarative input
+//     parsing + auto-generated OpenAPI) is poor fit here.
+//  2. The endpoint's contract is "redirect 302 with code+state in the
+//     URL query," not "JSON-in JSON-out." Huma's TokenOutput-style
+//     status+body shape isn't a clean way to model that.
+//
+// The downstream half (decode + consume) lives in
+// internal/service/oauth.go::authorizationCode and is served by
+// /oauth2/token.
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/rs/zerolog/log"
+
+	"github.com/highflame-ai/zeroid/internal/oautherror"
+	"github.com/highflame-ai/zeroid/internal/service"
+)
+
+// registerAuthorizeRoute mounts POST /oauth2/authorize on the public
+// chi router. GET is intentionally not supported in v1: RFC 6749
+// §4.1.1 permits GET for browser redirects, but the v1 use case (CLI
+// clients posting api_key + code_challenge) is POST-only. GET would
+// surface principal credentials in URL query strings + access logs.
+func (a *API) registerAuthorizeRoute(router chi.Router) {
+	router.Post("/oauth2/authorize", a.authorizeHandler)
+}
+
+// authorizeHandler is the chi handler for POST /oauth2/authorize.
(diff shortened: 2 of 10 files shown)


saucam:
Auto-fixed by pr-shepherd (iteration 1):
- **highflame-lint-check**: ran `gofmt -w tests/integration/oauth_authorize_test.go` (stripped one trailing blank line at EOF). Amended into the original `feat:` commit and force-pushed with lease — single-commit, single-author branch so amend was cleaner than a noise fixup commit.

**Stopping here — non-mechanical failure on `highflame-integration-test`.**

Every `TestAuthorize_*` subtest is failing with the same shape:
```
expected: 302 (or "invalid_request" body)
actual  : 400 + {"internalCode":"ZEROID-40001","message":"Content-Type must be application/json, got: application/x-www-form-urlencoded"}
```

Affected tests (all in `tests/integration/oauth_authorize_test.go`):
- TestAuthorize_HappyPath, TestAuthorize_StateOmitted, TestAuthorize_PKCERoundTrip, TestAuthorize_IssuedCodeShapeMatchesDecoder, TestAuthorize_RedirectURILoopbackEquivalence (expected 302, got 400)
- TestAuthorize_MissingRequiredFields/* (expected `invalid_request` OAuth error body, got JSON-content-type rejection)
- TestAuthorize_NonS256ChallengeMethod, TestAuthorize_NonCodeResponseType, TestAuthorize_RedirectURIMismatch (same)
- TestAuthorize_UnknownClient, TestAuthorize_NoPrincipalMatched, TestAuthorize_ResolverError (expected 401, got 400)

**Diagnosis** (for human triage, not shepherd-applied): a global content-type middleware in zeroid rejects non-`application/json` requests with the `ZEROID-40001` error before `handleAuthorize` runs. The new `/oauth2/authorize` route needs to either be exempted from that middleware (it's a browser-redirect endpoint where the params are query-string / form-encoded, not JSON), or the handler chain needs adjustment. No existing tests regressed — only the newly added authorize tests fail.

Other CI checks all green (build, docker, trivy, SDK smoke x2, socket security x2). No Gemini comments yet on any of the three surfaces.

Re-running CI on the gofmt fix to confirm lint passes. Re-shepherding the integration test is on you.

saucam:
## Re: Gemini's 6 review comments — 5 applied, 1 covered upstream (a2ff245)

@gemini-code-assist Thanks — the HIGH catches were both real and led to a coherent fix. Applied in commit a2ff245:

| # | File:Line | Suggestion | Disposition |
|---|-----------|------------|-------------|
| 1 | `internal/handler/authorize.go:166` | dead `a.resolvePrincipal == nil` branch | **APPLIED** — branch removed; replaced with `errors.Is(err, service.ErrNoResolversRegistered) → 503` |
| 2 | `server.go:615` | return `ErrNoResolversRegistered` from `resolvePrincipal` | **APPLIED** — new check `if len(resolvers) == 0` returns the sentinel |
| 3 | `internal/handler/authorize.go:75` | wrap `r.Body` with `http.MaxBytesReader` before `ParseForm` | **NOT APPLIED — already covered** — the upstream `requestValidationMiddleware` applies the 10 MiB MaxBytesReader to every request unconditionally (server.go:1175). My new `nonJSONContentTypeExemptPaths` only exempts the Content-Type check, not the body-size cap, so `/oauth2/authorize` retains the DoS guard you wanted. Adding it again in the handler would be double-coverage. |
| 4 | `internal/handler/authorize.go:31` | add `"errors"` import | **APPLIED** |
| 5 | `internal/service/principal.go:123` | define `ErrNoResolversRegistered` sentinel | **APPLIED** |
| 6 | `hooks.go:72` | re-export sentinel at top-level | **APPLIED** as `var ErrNoResolversRegistered = service.ErrNoResolversRegistered` |

Bonus fix in the same commit (not a Gemini comment but caught by the same CI run): every integration test was failing because `requestValidationMiddleware` was rejecting form-encoded POSTs before they reached my chi handler. Added `/oauth2/authorize` to `nonJSONContentTypeExemptPaths` — the body-size cap still applies to that path, only the Content-Type enforcement is exempted. Same theme as your resolver-chain catches: gaps in the request preflight that the test surface exposed.

rsharath (approved):

PR #179, opened 2026-06-03
feat: expose Server.ResolveAPIKey + APIKeyResolution type alias

saucam:
## Summary

Closes the consumer-side gap revealed when wiring the authn-side `api_key` PrincipalResolver against [#177](https://github.com/highflame-ai/zeroid/pull/177): `OAuthService.ResolveAPIKey` was added as the public-from-service projection of the api_key lookup, but `OAuthService` is package-internal — the surface deployer code (highflame-authn et al.) can actually reach is `*Server`, and the wrapper was missing.

Without this, the new `PrincipalResolver` hook from #177 is technically callable but has no clean way for the resolver to actually resolve an api_key. Consumers would have to either fork the lookup logic or fall back to the ugly `httptest.NewRecorder` /oauth2/token roundtrip pattern the lift was meant to retire.

## Changes

```go
// hooks.go
type APIKeyResolution = service.APIKeyResolution

// server.go
func (s *Server) ResolveAPIKey(ctx context.Context, apiKey string) (*APIKeyResolution, error) {
    return s.oauthSvc.ResolveAPIKey(ctx, apiKey)
}
```

Plus two integration tests in `tests/integration/oauth_authorize_test.go`:

- `TestServerResolveAPIKey` — pins the happy-path projection (account_id, project_id, user_id, scopes, key_id) against a freshly-minted api_key (via the existing `registerAgent` helper).
- `TestServerResolveAPIKey_UnknownKey` — pins the rejection-propagation contract (no swallowing).

No new test infrastructure needed — reuses the `testZeroIDServer` + `registerAgent` already in `helpers_test.go`.

## Why this didn't surface in #177's own tests

#177's integration tests use a stub `PrincipalResolver` that reads magic `test_principal_*` form fields and returns a pre-built `*zeroid.Principal` directly — it never actually calls `ResolveAPIKey`. The gap was invisible until I tried to wire a real consumer (the authn `APIKeyResolver`) against the surface.

Local validation pattern I now use to catch this class of bug **before** pushing:

1. Add a `replace github.com/highflame-ai/zeroid => /home/ydatta/Workspace/zeroid` in the downstream go.mod
2. Compile + test the downstream consumer against the local zeroid
3. Iterate on the zeroid surface until the consumer compiles cleanly
4. Push the zeroid PR with the consumer-verified surface

Saved as a workflow memory so I don't repeat the mistake.

## Verification

- `go vet ./...` and `go test ./internal/service/...` clean locally
- Integration tests verified to exercise the wrapper (require Postgres; CI runs them)
- End-to-end: `highflame-authn`'s rewrite of #93 compiles + passes all unit tests against this branch via local replace

## Test plan

- [ ] CI green
- [ ] No new Gemini comments warrant change (the wrapper is small + idiomatic)
- [ ] Spot-check: a downstream consumer's `srv.ResolveAPIKey(ctx, key)` call resolves through the public API surface

### DIFF
--- a/hooks.go
+++ b/hooks.go
@@ -82,6 +82,16 @@ var ErrPrincipalNotApplicable = service.ErrPrincipalNotApplicable
 // re-export exists so deployer tests can match on it via errors.Is.
 var ErrNoResolversRegistered = service.ErrNoResolversRegistered
 
+// APIKeyResolution is the public projection returned by
+// Server.ResolveAPIKey. Narrow + stable — does not leak zeroid
+// internals (*domain.Identity, *domain.APIKey row, credential-policy
+// records). Consumers map this onto whatever shape their layer needs
+// (typically zeroid.Principal in a PrincipalResolver implementation).
+//
+// See internal/service/oauth.go (APIKeyResolution definition) for the
+// canonical doc comment + field-level semantics.
+type APIKeyResolution = service.APIKeyResolution
+
 // AdminAuthMiddleware is an optional middleware applied to the admin API router.
 // When set, every request to the admin port passes through this middleware before
 // reaching any handler. Use this to add authentication (Bearer JWT, mTLS, API key,
--- a/internal/service/oauth.go
+++ b/internal/service/oauth.go
@@ -776,8 +776,11 @@ type APIKeyResolution struct {
 	// the developer whose CLI is being used right now. Empty for keys
 	// minted programmatically without a creator.
 	UserID string
-	// Scopes is the api key's intrinsic scope set. Empty means "no
-	// per-key restriction"; downstream policy may still narrow.
+	// Scopes is the api key's intrinsic scope set. Always non-nil
+	// (possibly len 0). Empty means "no per-key restriction";
+	// downstream policy may still narrow. Non-nil contract is to
+	// keep consumer code (range/len) safe without nil-checking and
+	// to JSON-marshal as `[]` rather than `null`.
 	Scopes []string
 	// KeyID is the API key's row UUID. Useful for audit/log
 	// attribution — never returned to end users.
@@ -885,8 +888,13 @@ func (s *OAuthService) resolveAPIKeyContext(ctx context.Context, apiKey string) 
 			AccountID: sk.AccountID,
 			ProjectID: sk.ProjectID,
 			UserID:    sk.CreatedBy,
-			Scopes:    append([]string(nil), sk.Scopes...),
-			KeyID:     sk.ID,
+			// Always non-nil (possibly len 0) so consumers can range /
+			// len safely without nil-checking. JSON-marshals to `[]`
+			// instead of `null` — better API hygiene for downstream
+			// serializers. Defensive copy so caller mutation can't
+			// leak back into the api_key row's slice.
+			Scopes: append(make([]string, 0, len(sk.Scopes)), sk.Scopes...),
+			KeyID:  sk.ID,
 		},
 		APIKey:   sk,
 		Identity: identity,
--- a/server.go
+++ b/server.go
@@ -522,6 +522,49 @@ func (s *Server) RegisterGrant(name string, handler GrantHandler) {
 	})
 }
 
+// ResolveAPIKey looks up a zid_sk_* API key and returns the resolved
+// tenant + user context. Public wrapper over the internal
+// OAuthService.ResolveAPIKey — exposed at Server level so deployer-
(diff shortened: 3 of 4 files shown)


saucam:
**pr-shepherd iteration 1 — stopping, handing off**

Polled CI to terminal: 7 SUCCESS, 2 FAILURE.

**Triage:**

1. **`Python SDK smoke tests` — infra flake.** Docker Hub registry pull for `testcontainers/ryuk:0.8.1` returned 500 (`context deadline exceeded` against `registry-1.docker.io`). Pure transient registry outage; TypeScript SDK smoke (same workflow, same testcontainers stack) ran fine right after. Rerun the failed job and it should pass. No code change warranted.

2. **`highflame-integration-test` — real test failure in the new test added by this PR.** Not mechanical, surfacing to you.

   ```
   --- FAIL: TestServerResolveAPIKey (0.00s)
       oauth_authorize_test.go:399:
           Error: Expected value not to be nil.
           Test:  TestServerResolveAPIKey
           Messages: Scopes must be addressable (empty slice or populated, never nil-vs-missing)
   ```

   The new test asserts `res.Scopes != nil` for a key minted by `registerAgent` (no explicit scopes). The underlying `OAuthService.ResolveAPIKey` (added in #177) returns `Scopes: nil` rather than `Scopes: []string{}` in that case — a real contract mismatch between this PR's test and the existing #177 projection. Resolving needs your call:

   - **Option A** — fix the contract in `OAuthService.ResolveAPIKey` to normalize `nil → []string{}` so the wrapper's contract is "never nil." Touches code outside this PR's stated 3-line wrapper scope, but matches what the test says the contract should be.
   - **Option B** — relax the new test to allow `nil` scopes (or drop the `NotNil` assertion entirely). Keeps this PR scope-tight; documents that `Scopes` is "nil-or-populated" rather than "empty-or-populated."

   Worth noting: the local pre-push verification (`go test ./internal/...`) didn't exercise this — the new test lives in `tests/integration/`, not `internal/`. Running `go test ./tests/integration/... -run TestServerResolveAPIKey` would have caught it.

**Side observation (NOT this PR's problem, but visible in the same job log):** three `panic: runtime error: invalid memory address or nil pointer dereference` panics in huma/v2 `SchemaLinkTransformer.Transform` during DELETE responses on `TestRFC7592_S2_4_*` tests. Chi's `Recoverer` middleware catches them so those tests still PASS, but it's a latent issue worth a separate ticket — pre-existing on main, completely unrelated to this PR.

**Gemini:** review landed at 02:52:33Z. Body: "No review comments were provided, so there is no additional feedback to address." Nothing to triage; no gemini-reviewer handoff needed.

**Suggested next step:** decide A vs B on `TestServerResolveAPIKey`, push the fix, rerun the Python SDK smoke job (or push triggers a new run automatically).

saucam:
## Update: fix pushed in 636603b — local validation now run end-to-end

Addressed both failures from the shepherd report:

**1. `TestServerResolveAPIKey` failure** — real contract bug in `OAuthService.ResolveAPIKey`. The test asserted `NotNil(res.Scopes)`; the implementation returned nil when the api_key had no scopes (`append([]string(nil), sk.Scopes...)` is nil-preserving). The test pinned the right production-grade shape — slice fields should be addressable, range/len safely, JSON-marshal to `[]` not `null`. Tightened the implementation to `append(make([]string, 0, len(sk.Scopes)), sk.Scopes...)` so the resolution is always non-nil. Doc comment updated.

**2. `Python SDK smoke tests` Docker Hub 500** — transient infra flake (testcontainers couldn't pull `testcontainers/ryuk:0.8.1`, TypeScript SDK ran clean in the same workflow). The push above triggers a fresh CI run automatically.

**Local validation gap closed for this turn:** previously ran `go test ./internal/...` (clean) but skipped `./tests/integration/...` because they need Postgres. Now ran `GOTOOLCHAIN=auto go test -run "TestAuthorize_|TestServerResolveAPIKey" ./tests/integration/ -count=1` against a local testcontainers Postgres — both `TestServerResolveAPIKey` and the existing `TestAuthorize_*` suite pass with the fix applied. Captured the discipline in my session memory so I don't skip integration tests again on PRs that add DB-touching public surface.

Also captured the unrelated huma `SchemaLinkTransformer.Transform` nil panics in the `TestRFC7592_S2_4_*` DELETE response path that the shepherd noticed in the integration log — chi's Recoverer catches them so the tests still pass, but worth a follow-up ticket since it's leaking panics into logs.

rsharath (approved):

akhiljavelin (approved):

PR #195, opened 2026-06-08
fix: accept any port on loopback redirect URIs per RFC 8252 §7.3

saucam:
Fixes the Overwatch PKCE failure: `redirect_uri is not in the client's registered list`.

## Root cause

The embedded Overwatch/ramparts **Guardian binds an ephemeral, OS-assigned loopback port** for its OAuth callback (`packages/guardian/src/module.ts` → `new GuardianHttpServer(config.httpPort || 0)` → port `0`). The redirect_uri it sends is therefore `http://127.0.0.1:<ephemeral>/oauth/callback`. `redirectURIAllowed` did an **exact-match including the port** against the fixed `overwatch-cli` seed list (`17580–17584`), so the ephemeral port was rejected.

This regressed when PKCE redirect validation **moved from Studio to AuthN**. Studio's `isValidRedirectUri` validated with a regex — `/^http:\/\/(127\.0\.0\.1|localhost):\d{4,5}\/oauth\/callback$/` — that accepts **any 4–5 digit loopback port**, so the ephemeral port passed pre-migration. AuthN's exact-match allow-list did not. (The regression test never caught it because it pinned a registered port, `127.0.0.1:17580`.)

## Fix

For **loopback** candidates (`127.0.0.1` / `localhost` / `::1`), match on scheme + host + path + query and **ignore the port**, per **RFC 8252 §7.3** (*"the authorization server MUST allow any port ... for loopback IP redirect URIs"*). Non-loopback URIs keep exact matching — no port leniency. Userinfo and fragments are rejected, and a look-alike like `127.0.0.1.evil.com` is not loopback, so the envelope matches the old anchored Studio regex.

The token-exchange identity check (RFC 6749 §4.1.3, `/oauth2/token`) is **unchanged** — it still compares the two client-supplied redirect_uris exactly, so the code stays bound to the specific ephemeral port issued at authorize time.

## Security

Reviewed by `security-reviewer` — **CLEAN, no bypass**. Empirically verified rejection of decimal/hex/octal IP encodings, `::ffff:127.0.0.1`, `evil@127.0.0.1`, `127.0.0.1@evil.com`, `127.0.0.1.evil.com`, trailing-dot, fragment/query injection, and backslash hosts; non-loopback hosts get zero port leniency; the emitted `Location` is built from the validated loopback candidate so a code can only ever reach a same-host listener (and PKCE binds it to the verifier regardless).

## Tests

`TestRedirectURIAllowed` gains loopback any-port cases (different / ephemeral / no port) plus negative guards (wrong path/scheme, userinfo, query, look-alike host, and non-loopback ports get **no** leniency). Service unit + OAuth integration tests green; lint 0 issues.

Paired with: highflame-regression#99 (de-fangs the masking test to use a non-registered loopback port).

### DIFF
--- a/internal/service/authcode_issue_test.go
+++ b/internal/service/authcode_issue_test.go
@@ -269,8 +269,23 @@ func TestRedirectURIAllowed(t *testing.T) {
 		{"exact match localhost", "http://localhost:17580/callback", true},
 		{"127.0.0.1 ↔ localhost equiv", "http://127.0.0.1:17580/callback", true},
 		{"exact match https", "https://app.example.com/oauth/callback", true},
-		{"wrong port", "http://localhost:17581/callback", false},
-		{"wrong path", "http://localhost:17580/other", false},
+
+		// RFC 8252 §7.3: any port is accepted on a loopback redirect, because
+		// the native/CLI app binds an ephemeral OS-assigned port.
+		{"loopback different port", "http://localhost:17581/callback", true},
+		{"loopback ephemeral high port", "http://localhost:54321/callback", true},
+		{"loopback 127.0.0.1 different port", "http://127.0.0.1:49999/callback", true},
+		{"loopback no port", "http://localhost/callback", true},
+
+		// Only the port floats — scheme / path / host / extras must still match.
+		{"loopback wrong path", "http://localhost:17580/other", false},
+		{"loopback wrong scheme", "https://localhost:17580/callback", false},
+		{"loopback with userinfo rejected", "http://evil@127.0.0.1:17580/callback", false},
+		{"loopback with query rejected", "http://localhost:17580/callback?x=1", false},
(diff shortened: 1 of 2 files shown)


abhijitjavelin (approved):

saucam:
pr-shepherd: CI is green — no fixes were required.

All 9 checks passed (build, lint, docker, trivy, Socket Security x2, Python + TypeScript SDK smoke, and the containerized OAuth/PKCE integration suite). No mechanical failures (commit prefix valid \`fix:\`, gofmt/goimports clean, go.mod/go.sum in sync, no Trivy CVEs). Branch was already up to date with main — no rebase, no force-push, no changes made to your branch.

Gemini review came back clean (no comments to address). Nothing left in-flight; this is ready for human review/merge.

PR #198, opened 2026-06-09
fix: security-review backlog — endpoint/grant client auth, policy ceiling, CIBA, SSRF, sweeps, config hardening

### LINKED ISSUE #201: OAuth introspection & revocation endpoints are unauthenticated (RFC 7662 §2.1 / RFC 7009)
## Problem

`/oauth2/token/introspect` and `/oauth2/token/revoke` accept **any** caller. They are registered in `registerOAuthRoutes` (`internal/handler/oauth.go:156-171`) with no `Security` and no auth middleware, and the handlers derive the tenant from the **submitted token itself** rather than from an authenticated caller identity:

- `introspectOp` — `internal/handler/oauth.go:328-334`
- `revokeOp` — `internal/handler/oauth.go:337-340`

[RFC 7662 §2.1](https://datatracker.ietf.org/doc/html/rfc7662#section-2.1): *"To prevent token scanning attacks, the endpoint MUST also require some form of authorization to access this endpoint."* An open introspection endpoint lets anyone probe token validity and metadata (active status, scopes, `cnf`, expiry) by submitting tokens — a token-scanning / validity-oracle surface.

For revocation, [RFC 7009 §2.1](https://datatracker.ietf.org/doc/html/rfc7009#section-2.1) requires client authentication; an unauthenticated `/revoke` lets anyone who can present a token force its revocation. Because revoke always returns `200` per RFC 7009 §2.2, it's also a blind operation with no caller accountability.

## Impact

- **Introspection:** token-scanning / validity oracle for anyone who obtains or guesses a token.
- **Revocation:** anyone holding a token can force-revoke it — an availability/DoS surface against agent credentials.
- **No caller-identity audit trail** on either endpoint; the tenant comes from the token, not the requester.

## Proposed change

Require caller authentication on both endpoints (RFC 7662 §2.1 / RFC 7009 §2.1):

- Confidential clients authenticate (client_secret / `private_key_jwt`, or the internal-service secret), consistent with how the rest of the platform authenticates service callers.
- Public clients: per RFC 7662, still require *some* authorization — decide the policy (e.g. restrict introspection to the internal mesh, or require a client credential alongside the token).
- Keep the RFC 7009 §2.2 "always 200" response for revoke, but gate access first.
- Derive/verify the tenant from the authenticated caller and cross-check it against the token's tenant, rather than trusting the token alone.

## References

- RFC 7662 §2.1 — introspection endpoint MUST require authorization (token-scanning defense)
- RFC 7009 §2.1 — revocation client authentication
- Code: `internal/handler/oauth.go:156-171` (registration), `:328-340` (handlers)
- Reported by Andrii Deinega via WIMSE WG review of ZeroID. Related: #193 (DPoP validation on the sibling `/oauth2/token/verify` forward-auth endpoint).

rsharath:
## Summary

Closes the medium-severity backlog from the deep security review (the critical cluster shipped in #197 as well as Closes #201). Built as six disjoint-file workstreams, integrated, and reviewed end-to-end before this PR. Full suite green under \`-race\` (696 tests); \`golangci-lint\` clean.

## What's fixed

**OAuth grant & endpoint client auth** (\`internal/service/oauth.go\`, \`internal/handler/oauth.go\`, \`internal/service/refresh_token.go\`)
- Confidential clients must now present \`client_secret\` on the **refresh_token** and **authorization_code** grants (RFC 6749 §6/§4.1.3/§10.4). Public/PKCE clients unchanged.
- **Introspection** (RFC 7662) and **revocation** (RFC 7009) now *verify* client credentials when presented (wrong secret → \`invalid_client\`); the existing internal/network-isolated path (no credentials) is preserved — non-breaking for that deployment model.
- **Scope-less client_credentials clients can no longer widen** — an empty registered scope set is deny, not allow-all (RFC 6749 §3.3).
- **Refresh-rotation no longer bricks a session.** A non-consuming \`PeekRefreshToken\` runs the client_id binding + identity-status gates *before* rotation consumes the token, so a transient failure (wrong client_id, briefly-suspended identity, DB blip) no longer trips reuse-detection and revokes the family. Unresolvable \`client_id\` fails closed (\`invalid_client\`) rather than skipping confidential-client auth.

**Policy ceiling** (\`internal/service/credential.go\`, \`internal/handler/credential.go\`, \`internal/service/attestation.go\`)
- The identity-policy ceiling (TTL / grant-type / delegation depth / attestation / scope) is now enforced on the three issuance paths that previously bypassed it: \`RotateCredential\`, the admin issue handler, and post-attestation issuance. Implemented via an opt-in \`IssueRequest.ResolveIdentityPolicy\` so the client-mediated grant paths (authcode/refresh/CIBA with no linked identity) keep their existing client-gated behavior — no over-reach.

**CIBA hardening** (\`internal/service/backchannel.go\`, \`internal/store/postgres/backchannel_request.go\`)
- \`Redeem\` requires a non-empty \`client_id\` and always compares it to the row (omitting it no longer skips the ownership check).
- Approved-but-unredeemed requests are bounded by \`expires_at\` + a grace window (enforced consistently across \`Redeem\`, \`MarkIssued\`, and \`DeleteExpired\`) instead of being redeemable forever.
- \`bc-authorize\` requires confidential-client auth and rejects deactivated clients.

**Attestation OIDC SSRF** (\`internal/attestation/oidc.go\`, \`internal/attestation/policy.go\`)
- Issuer discovery **and** the discovery-provided \`jwks_uri\` fetch now reject private / loopback / link-local / CGN / cloud-metadata IPs, with a dial-time re-check (DNS-rebind defense). The loopback-over-HTTP carve-out is gated behind a new \`AllowPrivateIssuerEndpoints\` flag (default **false**).

**Unbounded table growth** (\`internal/worker/cleanup.go\`, \`internal/store/postgres/refresh_token.go\`, migration \`032\`)
- Expired refresh tokens and signing credentials are now swept; refresh-token deletion lags the reuse-detection grace window so forensics aren't lost. Adds \`refresh_tokens(expires_at)\` index.

**Config hardening** (\`config.go\`, \`zeroid.yaml\`)
- Unparseable bool/int/float env values now **fail loud** at startup (a typo'd \`ZEROID_ALLOW_UNSAFE_DEV_STUB=flase\` no longer silently keeps the stub on).
- \`attestation.allow_unsafe_dev_stub=true\` and the default \`issuer\`/\`wimse_domain\` are **rejected when \`server.env == production\`** (gives the previously-dead \`server.env\` key a job); dev/test keep working.
- \`hmac_secret\` gets a ≥32-byte validation when set; new env mappings for \`hmac_secret\`, \`trust_forwarded_headers\`, \`backchannel.allow_private_notification_endpoints\`, and the attestation SSRF flag; removed telemetry keys now caught by \`rejectRemovedKeys\`.

## ⚠️ Behavior changes deployers should audit

1. **Confidential clients must authenticate** on refresh_token / authorization_code / CIBA bc-authorize. A confidential client that previously relied on PKCE/possession alone will get \`invalid_client\` until it sends its secret.
2. **client_credentials clients registered with zero scopes** are now rejected (\`invalid_scope\`) instead of receiving whatever they request.
3. **CIBA token requests must include \`client_id\`** (per CIBA Core §11) — a poller that omitted it will now be rejected.
4. **Production deployments must set an explicit issuer + wimse_domain and disable the dev stub** (these were silently defaulting).

## Testing

New tests pin each fix and were verified to fail with the fix reverted: \`oauth_authz_hardening_test.go\`, \`policy_ceiling_test.go\`, \`ciba_hardening_test.go\`, \`cleanup_sweep_test.go\`, \`internal/attestation/oidc_ssrf_test.go\`, plus \`config_test.go\` cases. Full unit + integration suite passes under \`-race\` (696 tests); CI now runs unit tests with the race detector (from #197).

## Review

Ran security + correctness review passes over the full diff before opening. Findings addressed: refresh-grant fails closed on an unresolvable client_id (was a confidential-auth-skip edge); confirmed no secret-hash persistence/leak on the CIBA path; stale code comment refreshed. No critical/high findings outstanding.

### DIFF
--- a/internal/attestation/oidc.go
+++ b/internal/attestation/oidc.go
@@ -90,10 +90,11 @@ func NewOIDCVerifier(httpClient *http.Client) *OIDCVerifier {
 // It is the seam through which a dev/test deployment opts out of the guard;
 // production must leave it false. Returns the verifier for chaining.
 //
-// NOTE FOR INTEGRATORS: there is currently no config flag wiring this from
-// server startup. To gate it on operator config, add an attestation config
-// field (see report in the closing PR) and call SetAllowPrivate in
-// server.go's verifier construction. Until then the guard is always on.
+// Wired from operator config: server.go passes
+// cfg.Attestation.AllowPrivateIssuerEndpoints (env
+// ZEROID_ATTESTATION_ALLOW_PRIVATE_ISSUER_ENDPOINTS, default false) into both
+// this verifier and the PolicyService, so the write-time and verify-time
+// guards agree. The guard is on unless an operator explicitly opts out.
 func (v *OIDCVerifier) SetAllowPrivate(allow bool) *OIDCVerifier {
 	v.allowPrivate = allow
 	return v
--- a/internal/attestation/oidc.go
+++ b/internal/attestation/oidc.go
@@ -476,28 +476,41 @@ func ssrfGuardedTransport(allowPrivate func() bool) *http.Transport {
 	base := http.DefaultTransport.(*http.Transport).Clone()
 	dialer := &net.Dialer{Timeout: 10 * time.Second}
 	base.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
-		if !allowPrivate() {
-			host, _, err := net.SplitHostPort(addr)
-			if err != nil {
-				host = addr
+		if allowPrivate() {
+			return dialer.DialContext(ctx, network, addr)
+		}
+		host, port, err := net.SplitHostPort(addr)
+		if err != nil {
+			host, port = addr, ""
+		}
+		// IP-literal host: validate and dial it directly.
+		if ip := net.ParseIP(host); ip != nil {
+			if isBlockedIP(ip) {
+				return nil, ErrPrivateAttestationEndpoint
 			}
-			if ip := net.ParseIP(host); ip != nil {
-				if isBlockedIP(ip) {
-					return nil, ErrPrivateAttestationEndpoint
-				}
-			} else {
-				// addr carries an IP literal only when the URL host was one
-				// (handled above). For hostnames, re-resolve here so a DNS
-				// rebind between the pre-flight check and this dial is still
-				// caught. The dialer below performs its own resolution, so a
-				// sub-millisecond rebind window remains — accepted residual
-				// risk for a host-based guard.
-				if err := resolveAndCheckHost(ctx, host); err != nil {
-					return nil, err
-				}
+			return dialer.DialContext(ctx, network, addr)
+		}
+		// Hostname: resolve ONCE, validate every answer, then dial a
+		// validated IP literal — not the hostname — so the kernel connects to
+		// exactly the address we checked. Re-resolving and dialing the
+		// hostname again (letting net.Dialer resolve a third time) would leave
+		// a DNS-rebinding window between the check and the connect; pinning the
+		// dial to the validated IP closes it. TLS still verifies against the
+		// original hostname because the Transport sets ServerName from the URL,
+		// not from the dial address.
+		ips, err := lookupIPs(ctx, host)
+		if err != nil {
+			return nil, fmt.Errorf("attestation issuer host resolution failed: %w", err)
+		}
+		if len(ips) == 0 {
+			return nil, ErrPrivateAttestationEndpoint
+		}
+		for _, ip := range ips {
+			if isBlockedIP(ip) {
+				return nil, ErrPrivateAttestationEndpoint
 			}
 		}
-		return dialer.DialContext(ctx, network, addr)
+		return dialer.DialContext(ctx, network, net.JoinHostPort(ips[0].String(), port))
 	}
 	return base
 }
--- a/internal/attestation/oidc.go
+++ b/internal/attestation/oidc.go
@@ -486,9 +486,12 @@ func ssrfGuardedTransport(allowPrivate func() bool) *http.Transport {
 					return nil, ErrPrivateAttestationEndpoint
 				}
 			} else {
-				// Resolver may hand back an IP literal as addr after a CGI
-				// rewrite, but for hostnames re-resolve here so a rebind
-				// between pre-flight check and dial is still caught.
+				// addr carries an IP literal only when the URL host was one
+				// (handled above). For hostnames, re-resolve here so a DNS
+				// rebind between the pre-flight check and this dial is still
+				// caught. The dialer below performs its own resolution, so a
+				// sub-millisecond rebind window remains — accepted residual
+				// risk for a host-based guard.
 				if err := resolveAndCheckHost(ctx, host); err != nil {
 					return nil, err
 				}
--- a/internal/attestation/oidc.go
+++ b/internal/attestation/oidc.go
@@ -3,10 +3,13 @@ package attestation
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"maps"
+	"net"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 	"time"
@@ -43,20 +46,57 @@ type OIDCVerifier struct {
 	cache    *jwksCache
 	http     *http.Client
 	cacheTTL time.Duration
+
+	// allowPrivate disables the SSRF blocklist on issuer / jwks_uri hosts.
+	// Production default is false: both the discovery GET and the JWKS GET
+	// resolve their target host and reject if ANY resolved IP falls in a
+	// private / loopback / link-local / metadata / CGN / multicast /
+	// unspecified range, re-checked at dial time as DNS-rebinding defence.
+	// Set to true ONLY in test/dev contexts that point issuers at loopback
+	// (httptest.NewServer, http://localhost:port). See SetAllowPrivate.
+	allowPrivate bool
 }
 
 // NewOIDCVerifier creates a verifier with a shared JWKS cache. httpClient
 // is used for both OIDC discovery and JWKS fetches; passing a custom one
 // is useful in tests (httptest.NewServer has no DNS).
+//
+// SSRF guard: when httpClient is nil, the verifier builds a hardened client
+// whose dialer rejects connections to private / reserved IP ranges. When a
+// custom client is supplied (tests), the caller owns transport behaviour —
+// but the pre-fetch host resolution check in discoverJWKSURL / fetchJWKS
+// still applies unless allowPrivate is set via SetAllowPrivate.
+//
+// allowPrivate defaults to false (production-safe). A tenant-configured
+// issuer that resolves to an internal address is rejected before any byte
+// leaves this process. To run against loopback fixtures, call
+// v.SetAllowPrivate(true) after construction.
 func NewOIDCVerifier(httpClient *http.Client) *OIDCVerifier {
-	if httpClient == nil {
-		httpClient = &http.Client{Timeout: 10 * time.Second}
-	}
-	return &OIDCVerifier{
+	v := &OIDCVerifier{
 		cache:    newJWKSCache(),
-		http:     httpClient,
 		cacheTTL: 1 * time.Hour,
 	}
+	if httpClient == nil {
+		httpClient = &http.Client{
+			Timeout:   10 * time.Second,
+			Transport: ssrfGuardedTransport(func() bool { return v.allowPrivate }),
+		}
+	}
+	v.http = httpClient
+	return v
+}
+
+// SetAllowPrivate toggles the SSRF blocklist on issuer / jwks_uri hosts.
+// It is the seam through which a dev/test deployment opts out of the guard;
+// production must leave it false. Returns the verifier for chaining.
+//
+// NOTE FOR INTEGRATORS: there is currently no config flag wiring this from
+// server startup. To gate it on operator config, add an attestation config
+// field (see report in the closing PR) and call SetAllowPrivate in
+// server.go's verifier construction. Until then the guard is always on.
+func (v *OIDCVerifier) SetAllowPrivate(allow bool) *OIDCVerifier {
+	v.allowPrivate = allow
+	return v
 }
 
 // ProofType reports oidc_token. One OIDCVerifier per registry.
@@ -97,8 +137,10 @@ func (v *OIDCVerifier) Verify(ctx context.Context, record *domain.AttestationRec
 		return nil, fmt.Errorf("oidc verifier: issuer not in allowlist: %s", issuerClaim)
 	}
 
-	// Step 3: resolve JWKS (cached).
-	keySet, err := v.cache.get(ctx, v.http, matchedIssuer.URL, v.cacheTTL)
+	// Step 3: resolve JWKS (cached). The cache fetch path runs the SSRF
+	// blocklist on both the issuer host (discovery GET) and the
+	// discovery-provided jwks_uri host (JWKS GET) before any network call.
+	keySet, err := v.cache.get(ctx, v.http, matchedIssuer.URL, v.cacheTTL, v.allowPrivate)
 	if err != nil {
 		return nil, fmt.Errorf("oidc verifier: JWKS fetch failed: %w", err)
 	}
@@ -221,7 +263,7 @@ func newJWKSCache() *jwksCache {
 // when the entry is absent or stale. The two-phase locking pattern (map
 // mutex for entry lookup, then entry mutex for fetch) ensures cross-issuer
 // lookups don't serialize on a single slow upstream.
-func (c *jwksCache) get(ctx context.Context, httpClient *http.Client, issuerURL string, ttl time.Duration) (jwk.Set, error) {
+func (c *jwksCache) get(ctx context.Context, httpClient *http.Client, issuerURL string, ttl time.Duration, allowPrivate bool) (jwk.Set, error) {
 	c.mu.Lock()
 	entry, ok := c.entries[issuerURL]
 	if !ok {
@@ -237,7 +279,7 @@ func (c *jwksCache) get(ctx context.Context, httpClient *http.Client, issuerURL 
 		return entry.keys, nil
 	}
 
-	jwksURL, err := discoverJWKSURL(ctx, httpClient, issuerURL)
+	jwksURL, err := discoverJWKSURL(ctx, httpClient, issuerURL, allowPrivate)
 	if err != nil {
 		return nil, err
 	}
@@ -247,7 +289,7 @@ func (c *jwksCache) get(ctx context.Context, httpClient *http.Client, issuerURL 
 	// them into a Set. Bound the body at maxDiscoveryDocBytes for the same
 	// reason we do on the discovery endpoint — defend against an issuer
 	// streaming attacker-chosen volumes into our process.
-	keySet, err := fetchJWKS(ctx, httpClient, jwksURL)
+	keySet, err := fetchJWKS(ctx, httpClient, jwksURL, allowPrivate)
 	if err != nil {
 		return nil, err
 	}
@@ -259,7 +301,13 @@ func (c *jwksCache) get(ctx context.Context, httpClient *http.Client, issuerURL 
 
 // fetchJWKS performs a bounded HTTP GET against jwksURL and parses the
 // response as a jwk.Set. Replaces the v2 jwk.Fetch helper that v4 removed.
-func fetchJWKS(ctx context.Context, httpClient *http.Client, jwksURL string) (jwk.Set, error) {
+func fetchJWKS(ctx context.Context, httpClient *http.Client, jwksURL string, allowPrivate bool) (jwk.Set, error) {
+	// SSRF guard: the jwks_uri came from the (untrusted) discovery document
+	// and may point at a different host than the issuer — resolve and block
+	// internal targets before connecting. Re-checked at dial time too.
+	if err := assertURLHostAllowed(ctx, jwksURL, allowPrivate); err != nil {
+		return nil, fmt.Errorf("jwks_uri %s: %w", jwksURL, err)
+	}
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, jwksURL, nil)
 	if err != nil {
 		return nil, fmt.Errorf("build JWKS request: %w", err)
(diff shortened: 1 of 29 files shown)


internal/service/oauth.go
```diff
@@ -1475,19 +1500,91 @@ func (s *OAuthService) refreshToken(ctx context.Context, req TokenRequest) (*dom
...
+	} else if req.ClientID != "" {
+		// A client_id was named but the client can't be resolved (deleted, or a
+		// transient lookup error). Fail closed: proceeding here would skip
+		// verifyConfidentialClientAuth entirely, letting a confidential client
+		// refresh with no secret after its registration is gone. The client_id
+		// binding check below also rejects any client_id that doesn't match the
+		// token's bound client, so a token genuinely issued without a client
+		// (req.ClientID == "") still refreshes via the no-op branch.
+		return nil, oauthBadRequest(oautherror.InvalidClient, "client_id does not resolve to a registered client")
 	}
```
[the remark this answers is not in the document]
rsharath: Deferring to the author. The fail-closed mapping here is deliberate and documented in the inline comment above (lines 1514-1521): on the refresh grant, any unresolvable client_id — deleted client OR transient lookup error — must NOT fall through to skip verifyConfidentialClientAuth, so it returns invalid_client. Your point that a transient DB error is arguably a 500 rather than a 400 is reasonable, but it changes the error contract on an auth path and is a security-judgment call, not a mechanical fix. Leaving it to the author/security-reviewer to decide whether to split ErrOAuthClientNotFound (400) from operational errors (500). pr-shepherd does not silently alter auth-path error semantics.

internal/service/oauth.go
```diff
@@ -1708,6 +1804,70 @@ func (s *OAuthService) parseWIMSEURI(wimseURI string) (accountID, projectID stri
...
+	}
+	if clientSecret == "" {
+		return oauthUnauthorized("client_secret is required for confidential clients", nil)
+	}
+	if _, err := s.oauthClientSvc.VerifyClientSecret(ctx, clientID, clientSecret); err != nil {
+		if errors.Is(err, ErrOAuthClientNotFound) || errors.Is(err, ErrInvalidClientSecret) {
+			return oauthUnauthorized("invalid client credentials", err)
+		}
+		return oauthUnauthorized("client verification failed", err)
+	}
```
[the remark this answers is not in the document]
rsharath: Deferring to the author. The code already separates the sentinel errors (ErrOAuthClientNotFound / ErrInvalidClientSecret -> 401) from the catch-all bucket; the open question is only whether the catch-all should be 401 (current, fail-closed on the confidential-client auth path) or 500 (oauthServerError). That is a deliberate error-contract decision on an authentication path, not a mechanical fix, so pr-shepherd is not flipping it autonomously. Flagging for the author / security-reviewer. (Note the same pattern also lives in VerifyPresentedClientAuth a few lines down — if you change one, change both for consistency.)

internal/worker/cleanup.go
```diff
@@ -141,6 +142,40 @@ func (w *CleanupWorker) RunOnce(ctx context.Context) {
...
+	// exactly (same WHERE: revoked = false AND audit_retention_until <= now) —
+	// replicated inline rather than wired through the repo because the worker is
+	// constructed in server.go without a SigningCredentialRepository and that
+	// constructor is out of scope to change. Revoked rows are retained as a
+	// tamper-evidence trail. Backed by idx_signing_credentials_retention (023).
+	signingRes, err := w.db.NewDelete().
+		TableExpr("signing_credentials").
+		Where("revoked = ?", false).
+		Where("audit_retention_until <= ?", now).
+		Exec(ctx)
```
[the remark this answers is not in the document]
rsharath: Applied in d054ae7. Switched to Model((*domain.SigningCredential)(nil)) — the model carries bun:"table:signing_credentials,alias:sc" so the table name and column mapping are now derived from the struct, matching SigningCredentialRepository.PruneExpiredRetention. Verified behavior is identical against a real Postgres (testcontainers): the typed delete reaps expired non-revoked credentials and retains both revoked (tamper trail) and in-retention-window rows. go build / go vet / golangci-lint clean.

internal/service/oauth.go
```diff
@@ -1708,6 +1804,70 @@ func (s *OAuthService) parseWIMSEURI(wimseURI string) (accountID, projectID stri
...
+		return oauthUnauthorized("client_id and client_secret must both be supplied", nil)
+	}
+	if _, err := s.oauthClientSvc.VerifyClientSecret(ctx, clientID, clientSecret); err != nil {
+		if errors.Is(err, ErrOAuthClientNotFound) || errors.Is(err, ErrInvalidClientSecret) {
+			return oauthUnauthorized("invalid client credentials", err)
+		}
+		return oauthUnauthorized("client verification failed", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
rsharath: Applied in 0385842. Good catch — this was a real bug in the accept-and-verify logic. `VerifyPresentedClientAuth` now allows a registered **public** client to present only its `client_id` on introspect/revoke (RFC 7009 §2.1 / RFC 7662 §2.1 — public clients have no secret, and standard libraries auto-attach client_id to revocation requests). A confidential client_id without a secret, or an unknown client_id, is still rejected, so this neither widens access nor lets a confidential client skip auth. Added `TestOAuthAuthzHardening_PublicClientRevocationWithoutSecret` covering both branches.

internal/service/oauth.go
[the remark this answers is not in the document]
rsharath: Applied in 0385842. The catch-all now returns `oauthServerError` (500) instead of 401 for a non-sentinel error. Note: in practice `VerifyClientSecret` collapses every failure to `ErrOAuthClientNotFound`/`ErrInvalidClientSecret` (it doesn't propagate the underlying repo error), so the catch-all is currently unreachable — but mapping it to 500 is the correct contract and future-proofs the path if that service ever surfaces operational errors.

internal/service/oauth.go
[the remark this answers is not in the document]
rsharath: Applied in 0385842. The refresh else-branch now distinguishes `ErrOAuthClientNotFound` (→ invalid_client) from any other error (→ oauthServerError/500), matching your suggestion. Caveat: `GetClientByClientID` currently collapses all errors (including transient DB errors) to `ErrOAuthClientNotFound`, so the 500 path won't fire until that lookup is refactored to propagate operational errors — a separate, broader change. The guard is in place for when it is.

internal/attestation/oidc.go
```diff
@@ -329,3 +385,172 @@ func discoverJWKSURL(ctx context.Context, httpClient *http.Client, issuerURL str
...
+				// Resolver may hand back an IP literal as addr after a CGI
+				// rewrite, but for hostnames re-resolve here so a rebind
+				// between pre-flight check and dial is still caught.
+				if err := resolveAndCheckHost(ctx, host); err != nil {
+					return nil, err
+				}
+			}
+		}
+		return dialer.DialContext(ctx, network, addr)
+	}
```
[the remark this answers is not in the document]
rsharath: Applied in 19800e8. `ssrfGuardedTransport` now resolves the host once, validates every returned IP, and dials the validated IP literal via `net.JoinHostPort(ips[0], port)` rather than re-dialing the hostname — so the kernel connects to exactly the address checked, closing the rebind window. TLS still verifies against the original hostname (the Transport sets ServerName from the URL, not the dial address).

internal/store/postgres/refresh_token.go
```diff
@@ -108,6 +108,28 @@ func (r *RefreshTokenRepository) ClaimByTokenHash(ctx context.Context, db bun.ID
...
+		Model((*domain.RefreshToken)(nil)).
+		Where("expires_at < ?", before).
+		Exec(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("failed to delete expired refresh tokens: %w", err)
+	}
+
+	n, _ := res.RowsAffected()
+
+	return n, nil
```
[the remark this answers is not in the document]
rsharath: Applied in 19800e8. `DeleteExpired` now returns the `RowsAffected` error instead of discarding it, matching `ClaimByTokenHash` and the other repo methods.

internal/handler/oauth.go
```diff
@@ -326,17 +343,37 @@ func (a *API) tokenOp(ctx context.Context, input *TokenInput) (*TokenOutput, err
...
+	// Accept-and-verify client auth (RFC 7009 §2.1). A presented-but-wrong
+	// secret is rejected with invalid_client; otherwise the RFC 7009 §2.2
+	// "always 200" contract holds (including for unknown/already-revoked tokens
+	// and anonymous internal-path callers).
+	if err := a.oauthSvc.VerifyPresentedClientAuth(ctx, input.Body.ClientID, input.Body.ClientSecret); err != nil {
+		_, desc, status := extractOAuthError(err)
+		out := &OAuthRevokeOutput{Status: status}
+		out.Body.Error = oautherror.InvalidClient
+		out.Body.ErrorDescription = desc
+		return out, nil
```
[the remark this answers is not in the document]
rsharath: Applied in 19800e8. `revokeOp` now uses the code from `extractOAuthError` instead of hardcoding `invalid_client`, so an operational 500 surfaces as server_error rather than a mislabelled client-auth failure. (`introspectOp` already did this correctly.)

internal/service/oauth.go
```diff
@@ -1708,6 +1808,89 @@ func (s *OAuthService) parseWIMSEURI(wimseURI string) (accountID, projectID stri
...
+	if _, err := s.oauthClientSvc.VerifyClientSecret(ctx, clientID, clientSecret); err != nil {
+		if errors.Is(err, ErrOAuthClientNotFound) || errors.Is(err, ErrInvalidClientSecret) {
+			return oauthUnauthorized("invalid client credentials", err)
+		}
+		// An unexpected (operational) error is a server fault, not a client
+		// authentication failure — don't mislead the client with a 401.
+		return oauthServerError("client verification failed", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
rsharath: Applied in 19800e8. `verifyConfidentialClientAuth` now treats a client with a non-empty stored secret hash as confidential even if `client_type` isn't "confidential" — the same belt-and-suspenders test the CIBA bc-authorize/redeem paths use — so an inconsistent row can't skip secret verification.

tests/integration/cleanup_sweep_test.go
```diff
@@ -0,0 +1,125 @@
...
+	// grace; this row's expires_at is now - grace/2 > cutoff) → RETAIN. This is
+	// the reuse-forensics guard: a row whose expiry is more recent than the
+	// grace lag must survive the sweep.
+	recentRevokedAt := now.Add(-grace / 2)
+	withinGrace := insertRefreshTokenRow(t,
+		uid("rt-within-grace"),
+		domain.RefreshTokenStateRevoked,
+		now.Add(-grace/2),
+		&recentRevokedAt,
+	)
```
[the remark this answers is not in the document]
rsharath: Applied in 19800e8. The within-grace row now uses a fixed `now - 1s` expiry instead of `now - grace/2`, giving ~grace-1s (≈9s) of margin to the worker's cutoff — well beyond any plausible scheduling delay between capturing the reference time and the cleanup run.

internal/handler/oauth.go
adeinega, line 1: It’s worth mentioning that `ClientSecret` in struct `Body` of `IntrospectInput` and `OAuthRevokeInput` covers only the client_secret_post authentication method.

There are other popular ones such as `client_secret_basic`, `private_key_jwt` and wonderful `tls_client_auth`, and also less popular `client_secret_jwt`.

OAuth 2.1 effectively dropped the latter one (`client_secret_jwt`).

adeinega: The Authorization Server Metadata document can also be updated with two properties `introspection_endpoint_auth_methods_supported ` and `revocation_endpoint_auth_methods_supported`.

internal/handler/oauth.go
```diff
@@ -113,22 +113,39 @@ func extractOAuthError(err error) (code, description string, status int) {
...
+		ClientSecret string `json:"client_secret,omitempty" doc:"OAuth client secret (optional client auth)"`
 	}
 }
 
 type OAuthRevokeOutput struct {
-	Body struct {
+	Status int
+	Body   struct {
 		Revoked bool `json:"revoked"`
+		// Error / ErrorDescription are populated only when presented client
```
adeinega, line 160: A complaint revocation endpoint literally returns a 200 HTTP status code (OK) regardless of whether the client provided a valid or invalid token per https://datatracker.ietf.org/doc/html/rfc7009#section-2.2.

> Note: invalid tokens do not cause an error response since the client
> cannot handle such an error in a reasonable way.  Moreover, the
> purpose of the revocation request, invalidating the particular token,
> is already achieved.

The spec doesn’t define any other properties.

internal/handler/oauth.go
rsharath: Applied in 24f6771. Both endpoints now accept `client_secret_basic` via the Authorization header (RFC 6749 Section 2.3.1, form-urlencoded credentials) alongside the existing `client_secret_post` body fields. Presenting both methods in one request is rejected with `invalid_request` per Section  2.3, and a failed Basic attempt gets the Section 5.2 `WWW-Authenticate: Basic` echo on the 401.

`private_key_jwt` and `tls_client_auth` are real features — leaving those as follow-ups. See https://github.com/highflame-ai/zeroid/issues/206 (opened to track this) and `client_secret_jwt` deliberately skipped since OAuth 2.1 dropped it, as you note.

rsharath: Applied in 24f6771. The RFC 8414 document now advertises `introspection_endpoint_auth_methods_supported` and `revocation_endpoint_auth_methods_supported`, both as `["client_secret_post", "client_secret_basic"]` — matching what the endpoints actually accept after the change above. Pinned by an integration test against /.well-known/oauth-authorization-server.

internal/handler/oauth.go
rsharath: The 200-always contract already holds here for token outcomes: `revokeOp` discards the revoke result and returns 200 for valid, invalid, unknown, and already-revoked tokens alike — exactly the Section 2.2 note you quote. The only non-200 paths are failed *client authentication* (the Section 2.2.1 carve-out, error per RFC 6749 Section 5.2 — now including the Basic 401 + WWW-Authenticate echo) and a 500 on operational failure.

On "the spec doesn't define any other properties": agreed the `{"revoked": true}` body is a ZeroID extension. Section 2.2 says the response body is ignored by compliant clients, so it should be interop-safe — and the published Rust SDK hard-deserializes `TokenRevoked { revoked: bool }` (TS parses it too), so removing it would break released clients on an empty body. Decided to strip it in a coordinated SDK major release than in this PR. Keeping it for now, documented as an extension.

internal/handler/oauth.go
rsharath: Follow-up filed as #206. Investigating it turned up something worth knowing: registration already accepts `token_endpoint_auth_method: private_key_jwt` and stores `jwks`/`jwks_uri`, but nothing enforces it at token time — a private_key_jwt client with no secret is currently treated as a *public* client (no auth at all). So the issue is scoped as closing an advertised-but-unenforced gap, not just adding a method. `tls_client_auth` is explicitly deferred in the issue pending an mTLS termination story / X.509 SVID work.

internal/service/backchannel.go
```diff
@@ -353,15 +362,47 @@ func (s *BackchannelService) CreateAuthRequest(ctx context.Context, in CreateAut
...
+	// initiate against a confidential client lets them spam prompts at
+	// arbitrary users under that client's identity. Require + verify the
+	// client_secret when the client is confidential. A client is confidential
+	// if it declares so or carries a stored secret hash (belt-and-suspenders).
+	if client.ClientType == "confidential" || client.ClientSecret != "" {
+		if in.ClientSecret == "" {
+			return nil, oauthBadRequest(oautherror.InvalidClient, "client_secret is required for a confidential client")
+		}
+		// VerifyClientSecret re-loads by client_id, re-checks is_active, and
+		// compares the bcrypt hash. On any failure surface the opaque
```
adeinega, line 398: On a side note, https://pkg.go.dev/crypto/pbkdf2 is a better choice than bcrypt in case one day you get a requirement to become a FIPS compliant app.

internal/handler/oauth.go
```diff
@@ -111,27 +114,99 @@ func extractOAuthError(err error) (code, description string, status int) {
...
+// methods on the introspection/revocation endpoints: client_secret_basic
+// (Authorization header) and client_secret_post (body fields). Returns the
+// effective credentials plus whether the Basic header was the source (drives
+// the RFC 6749 §5.2 WWW-Authenticate echo on failure).
+//
+// Per RFC 6749 §2.3 a client MUST NOT use more than one method in a single
+// request — presenting both is rejected as invalid_request. Non-Basic
+// Authorization schemes are ignored (treated as not presented) so bearer
+// headers from generic middleware don't break the anonymous internal path.
+// Credentials inside Basic are form-urlencoded per RFC 6749 §2.3.1.
```
adeinega, line 182: 👍 

The absence of `url.QueryUnescape(id)` and `url.QueryUnescape(secret)` has resulted in numerous integration issues (in other implementations).

rsharath: Thanks — yes, that was deliberate. `resolveInspectionClientAuth` runs `url.QueryUnescape` on both halves after splitting on the first `:` (RFC 6749 §2.3.1 / Appendix B: the id and secret are form-urlencoded before base64), so client_ids or secrets containing `:`, `+`, spaces, etc. round-trip correctly. Covered by the `TestInspectionBasicAuth_*` integration tests.

internal/service/backchannel.go
rsharath: Good call, and noted for the roadmap. It's a deliberate non-change for this PR though: bcrypt is the hash for every stored secret in ZeroID (oauth client secrets, API keys, DCR registration-access tokens — internal/service/oauth_client.go, backchannel.go, oauth.go, dynamic_registration.go), so swapping to PBKDF2 is a cross-cutting credential-format migration (rehash-on-verify or forced rotation), not a local edit. We don't have a FIPS requirement on the books yet, I'll file it as its own tracked piece of work covering the whole credential surface (and the Go `GODEBUG=fips140` / boringcrypto build story too).

rsharath: Filed as #207 to track this independently (credential-hashing migration off bcrypt → PBKDF2 + the FIPS-validated-module build story). Credited you there.

tests/integration/inspection_basic_auth_test.go
```diff
@@ -0,0 +1,150 @@
...
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	return decode(t, resp)["access_token"].(string), client
+}
+
+// TestInspectionBasicAuth_IntrospectAcceptsBasic verifies a valid
+// client_secret_basic header authenticates an introspection call.
+func TestInspectionBasicAuth_IntrospectAcceptsBasic(t *testing.T) {
+	token, client := inspectionFixture(t, "basic-introspect")
+
+	resp := post(t, "/oauth2/token/introspect", map[string]any{
```
adeinega, line 53: I'd suggest covering the same scenario using `postForm(...)` as well, for this and other similar tests in this file.

rsharath: Done in badce2c. Added form-encoded (`postForm`) twins of the meaningful cases: accept-Basic, wrong-secret → 401 + `WWW-Authenticate: Basic`, both-methods → `invalid_request`, and the revoke happy path. Good catch: the form-compat middleware and the Authorization-header binding are independent code paths, and only their intersection (Basic header + form body - the realistic stock-client shape) was untested before. I left the malformed-header and metadata cases JSON-only since they're content-type agnostic (header parsing is identical; metadata is a GET). All 11 TestInspectionBasicAuth_* pass.

saucam (approved):

PR #210, opened 2026-06-19
fix: sweep expired refresh_tokens to bound table growth

rsharath:
## Problem

`refresh_tokens` grew **unbounded**. Under rotation, every refresh inserts a successor row and revokes its predecessor (RFC 6749 §6 / reuse detection), but nothing ever deleted the old rows. The cleanup worker reaps `issued_credentials`, `proof_tokens`, `auth_codes`, `dpop_jti`, and CIBA requests — refresh tokens were the one expiring table left out.

## Change

- **Cleanup sweep** (`internal/worker/cleanup.go`): `DELETE FROM refresh_tokens WHERE expires_at < now`. An expired refresh token is unusable — every lookup (`GetByTokenHash` / `ClaimByTokenHash`) already requires `expires_at > now`.
  - **No retention window.** Unlike `issued_credentials`, whose expired rows hold the `parent_jti` edges the cascade-revocation walk still needs, `refresh_tokens` have no cross-row dependency. Revoked-but-*unexpired* rows are kept — they're the reuse-detection tripwires for still-active families; only past-expiry rows drop.
- **Migration 033** adds `idx_refresh_tokens_expires_at` so the sweep filters on an index instead of a full table scan, mirroring `idx_issued_credentials_expires_at` (migration 001). Plain `CREATE INDEX` — golang-migrate wraps each migration in a transaction, so `CONCURRENTLY` can't run there.

## Test

`tests/integration/refresh_token_cleanup_test.go` — asserts an expired row is swept and an unexpired one survives. Passes locally (`GOEXPERIMENT=jsonv2 go test`).

## Note

A duplicate, never-committed `032_refresh_tokens_expires_at_index` (collided with the committed `032_identity_status_expired` from #208, and indexed for a sweep that didn't exist) was discarded in favor of this `033` that ships the index **and** the sweep together.

### DIFF
--- a/internal/worker/cleanup.go
+++ b/internal/worker/cleanup.go
@@ -18,7 +18,8 @@ type IdentityExpirer interface {
 }
 
 // CleanupWorker periodically removes expired issued_credentials, proof_tokens,
-// and auth_codes rows, and sweeps expired identities into status=deactivated
+// auth_codes, refresh_tokens, and dpop_jti rows, and sweeps expired identities
+// into status=deactivated
 // via IdentityService.SweepExpiredIdentities (an atomic conditional UPDATE
 // claim followed by the existing runDeactivationCleanup cascade).
 // Running the cleanup prevents unbounded table growth since credentials have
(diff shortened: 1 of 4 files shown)


rsharath:
pr-shepherd stopped (iteration 1) — handing off to a human.

**Why I stopped:** This PR re-implements a feature that already merged to main, and the conflict is a deliberate design disagreement, not a mechanical one — outside my auto-fix scope.

**What I found:**
- Branch is 1 behind / 1 ahead of `main`. `main` tip `bf53b45` (merged PR #198, security-review backlog) **already added migration 033 (`idx_refresh_tokens_expires_at`) and a refresh_tokens expiry sweep in `internal/worker/cleanup.go`.**
- `git rebase origin/main` conflicts (add/add on `migrations/033_refresh_tokens_expires_at_index.up.sql`; `internal/worker/cleanup.go` also overlaps). I aborted the rebase per policy — I do not resolve semantic conflicts.
- The two sweep designs are mutually exclusive:
  - **main:** deletes `expires_at < now - RefreshTokenReuseGraceWindow` (lags the cutoff by the reuse-detection grace window, via `refreshRepo.DeleteExpired`).
  - **this PR:** deletes `expires_at < now` with **no grace window**, and also **removes** main's signing-credentials retention prune block + the `domain` import.
- Migration number 033 is already taken on main — this PR's 033 must be renumbered (034) regardless of which sweep wins.

**CI state:** The full Highflame CI suite (commit-check, go build, go test, Trivy) is **not running** — GitHub gates those workflows behind mergeability and the branch is `CONFLICTING`. Only Socket Security ran (1 pass, 0 fail). There are no failed checks for me to mechanically fix.

**Gemini:** 1 inline comment on `cleanup.go:143` (medium) — prefer Bun strongly-typed `.Model((*domain.RefreshToken)(nil))` over the string table name. Deferred: it targets the exact lines the conflict resolution will rewrite, and needs the design question settled first.

**Suggested next steps (human judgment required):**
1. Decide whether main's grace-window sweep already satisfies this PR's intent — if so, this PR may be redundant / should close.
2. If a no-grace-window sweep is genuinely wanted, that is a reuse-detection safety call (does dropping the grace window risk reaping rows reuse detection still needs?) — and confirm whether removing main's signing-credentials prune is intentional.
3. Renumber migration 033 → 034.
4. Then rebase on main; once it is mergeable, CI will run and I can take over any mechanical failures, and `gemini-reviewer` can triage the cleanup.go:143 comment.

I have not pushed anything. Working tree and branch are untouched (original commit `f7d8591`).

PR #225, opened 2026-07-07
fix: B1 — retain issued_credentials past expiry so the delegation graph survives

KunalJavelin:
Part of highflame-ai/highflame-architecture#132 (epic), bug #142 (B1).

## Problem
The delegation graph in Studio blanks out after a token expires. Root cause: `issued_credentials` rows were hard-pruned at (or shortly after) credential expiry, so the lineage the graph is built from (jti → parent_jti → mission_id) disappeared the moment a delegated token lapsed — exactly when you still want to inspect what happened.

## Fix
Retain `issued_credentials` **past expiry** so the delegation chain survives for audit/inspection, then prune on a separate, longer audit-retention clock:

- Two-clock retention on `issued_credentials` (migration): a row is only eligible for pruning when **both** its credential-expiry and its `audit_retention_until` have passed.
- `token.audit_retention_days` config (default 400, max 36500) as one source of truth (`domain.DefaultAuditRetentionDays` / `MaxAuditRetentionDays`), parsed correctly from env (was a real int-parse bug).
- Rotation-of-expired guard: rotating a dead credential returns a clean 409 (sentinels) instead of a 500.

## Tests
- [x] Migration lock-posture + fixed-400d backfill; prune requires both clocks; rotation-of-expired → 409.
- [x] Green on the zeroid suite (GOEXPERIMENT=jsonv2).

Follow-up: zeroid#226 (bounded graph root-expansion read path).

### DIFF
--- a/config.go
+++ b/config.go
@@ -236,6 +236,15 @@ type TokenConfig struct {
 	DefaultTTL int    `koanf:"default_ttl"`
 	MaxTTL     int    `koanf:"max_ttl"`
 
+	// AuditRetentionDays is how long an issued_credentials row remains
+	// queryable AFTER the token expires. The two clocks are deliberately
+	// decoupled (same model as signing_credentials): expires_at bounds how
+	// long the token authenticates; expires_at + AuditRetentionDays bounds
+	// how long the delegation graph can still answer historical questions
+	// about it. The cleanup worker prunes on the audit clock, never on
+	// expiry alone. See domain.IssuedCredential.AuditRetentionUntil.
+	AuditRetentionDays int `koanf:"audit_retention_days"`
+
 	// authorization_code grant configuration.
 	// HMACSecret is the shared secret used to sign and verify auth code JWTs (HS256).
 	HMACSecret string `koanf:"hmac_secret"`
@@ -554,6 +563,10 @@ func loadDefaults(k *koanf.Koanf) error {
 		// validateWIMSEDomain() reject empty values at startup.
 		"token.default_ttl": 3600,
 		"token.max_ttl":     7776000, // 90 days
+		// Delegation-graph evidence clock: issued_credentials rows stay
+		// queryable this long past token expiry (mirrors
+		// signing_credentials.audit_retention_days).
+		"token.audit_retention_days": 400,
 		// Accept-and-verify on introspect/revoke by default (dev/standalone).
 		// Validate() forces this false in production (RFC 7662/7009).
 		"token.allow_unauthenticated_token_inspection": true,
@@ -634,6 +647,8 @@ func loadEnvVars(k *koanf.Koanf) error {
 		"ZEROID_ISSUER":                "token.issuer",
 		"ZEROID_TOKEN_TTL_SECONDS":     "token.default_ttl",
 		"ZEROID_MAX_TOKEN_TTL_SECONDS": "token.max_ttl",
+		// Evidence clock for issued_credentials (delegation-graph retention).
+		"ZEROID_TOKEN_AUDIT_RETENTION_DAYS": "token.audit_retention_days",
 		// HMAC secret signs/verifies stateless authorization_code JWTs (HS256).
 		// A leak forges auth codes; Validate() enforces >= 32 bytes when set.
 		"ZEROID_HMAC_SECRET": "token.hmac_secret",
--- a/config.go
+++ b/config.go
@@ -390,6 +390,16 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("token.hmac_secret must be at least 32 bytes when set, got %d: it signs stateless auth-code JWTs (HS256) and a short secret is forgeable", len(c.Token.HMACSecret))
 	}
 
+	// token.audit_retention_days is a day count converted to time.Duration
+	// (days × 24h in nanoseconds); past ~106751 days the multiplication
+	// overflows int64 into a NEGATIVE duration, which would stamp
+	// AuditRetentionUntil before expiry and silently erase the evidence
+	// window the knob exists to guarantee. 36500 (~100y) is a generous
+	// practical ceiling. 0 means "use the shipped default (400)".
+	if c.Token.AuditRetentionDays < 0 || c.Token.AuditRetentionDays > 36500 {
+		return fmt.Errorf("token.audit_retention_days must be between 0 and 36500, got %d: it is the delegation-graph evidence-retention window in days (0 = default 400)", c.Token.AuditRetentionDays)
+	}
+
 	seen := make(map[string]struct{}, len(c.ExternalIssuers))
 	for i := range c.ExternalIssuers {
 		c.ExternalIssuers[i].Defaults()
--- a/config.go
+++ b/config.go
@@ -243,6 +243,11 @@ type TokenConfig struct {
 	// how long the delegation graph can still answer historical questions
 	// about it. The cleanup worker prunes on the audit clock, never on
 	// expiry alone. See domain.IssuedCredential.AuditRetentionUntil.
+	//
+	// Floor: the cleanup worker ANDs this evidence clock with the
+	// cascade-walk clock (Token.MaxTTL). A value below MaxTTL (~90d) has no
+	// effect on deletion timing — MaxTTL is the effective floor. 0 selects
+	// domain.DefaultAuditRetentionDays; see Validate for the accepted range.
 	AuditRetentionDays int `koanf:"audit_retention_days"`
 
 	// authorization_code grant configuration.
@@ -394,10 +399,10 @@ func (c *Config) Validate() error {
 	// (days × 24h in nanoseconds); past ~106751 days the multiplication
 	// overflows int64 into a NEGATIVE duration, which would stamp
 	// AuditRetentionUntil before expiry and silently erase the evidence
-	// window the knob exists to guarantee. 36500 (~100y) is a generous
-	// practical ceiling. 0 means "use the shipped default (400)".
-	if c.Token.AuditRetentionDays < 0 || c.Token.AuditRetentionDays > 36500 {
-		return fmt.Errorf("token.audit_retention_days must be between 0 and 36500, got %d: it is the delegation-graph evidence-retention window in days (0 = default 400)", c.Token.AuditRetentionDays)
+	// window the knob exists to guarantee. Bounds shared with the service
+	// clamp via domain. 0 means "use domain.DefaultAuditRetentionDays".
+	if c.Token.AuditRetentionDays < 0 || c.Token.AuditRetentionDays > domain.MaxAuditRetentionDays {
+		return fmt.Errorf("token.audit_retention_days must be between 0 and %d, got %d: it is the delegation-graph evidence-retention window in days (0 = default %d)", domain.MaxAuditRetentionDays, c.Token.AuditRetentionDays, domain.DefaultAuditRetentionDays)
 	}
 
 	seen := make(map[string]struct{}, len(c.ExternalIssuers))
@@ -574,9 +579,9 @@ func loadDefaults(k *koanf.Koanf) error {
 		"token.default_ttl": 3600,
 		"token.max_ttl":     7776000, // 90 days
 		// Delegation-graph evidence clock: issued_credentials rows stay
-		// queryable this long past token expiry (mirrors
-		// signing_credentials.audit_retention_days).
-		"token.audit_retention_days": 400,
+		// queryable this long past token expiry. Single source of truth in
+		// domain so the default can't drift from the service clamp / Validate.
+		"token.audit_retention_days": domain.DefaultAuditRetentionDays,
 		// Accept-and-verify on introspect/revoke by default (dev/standalone).
 		// Validate() forces this false in production (RFC 7662/7009).
 		"token.allow_unauthenticated_token_inspection": true,
@@ -714,6 +719,7 @@ func loadEnvVars(k *koanf.Koanf) error {
 			strings.HasSuffix(configPath, ".max_idle_conns") ||
 			strings.HasSuffix(configPath, ".default_ttl") ||
 			strings.HasSuffix(configPath, ".max_ttl") ||
+			strings.HasSuffix(configPath, ".audit_retention_days") ||
 			strings.HasSuffix(configPath, ".shutdown_timeout_seconds"):
 			intVal, err := strconv.Atoi(value)
 			if err != nil {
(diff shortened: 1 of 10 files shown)


internal/service/credential.go
```diff
@@ -668,6 +686,14 @@ func (s *CredentialService) RotateCredential(ctx context.Context, credID, accoun
 	if old.IsRevoked {
 		return nil, nil, fmt.Errorf("credential is already revoked")
 	}
+	// An expired credential cannot be rotated: the revoke half would be a
+	// silent no-op (the cascade anchor requires expires_at > revoked_at) and
+	// "rotate" would degrade to minting a fresh token off a dead row. Rows
+	// now outlive expiry by the audit-retention window, so without this
+	// guard a long-dead credential would stay rotatable for ~400 days.
+	if time.Now().After(old.ExpiresAt) {
+		return nil, nil, fmt.Errorf("credential is expired; issue a new credential instead of rotating")
```
saucam: This error surfaces to clients as a 500: `rotateCredentialOp` in `internal/handler/credential.go` maps every `RotateCredential` error to `huma.Error500InternalServerError`. Now that rows deliberately outlive their tokens by ~400 days, rotating a long-expired credential becomes a routine client mistake, and each one will fire 5xx alerting and tell the caller to retry something that can never succeed.

The revoke path already does this right (`revokeCredentialOp` branches on `ErrCredentialNotFound` to a 404). Suggest a sentinel error (e.g. `ErrCredentialExpired`) and a 409/410 mapping in the handler. Same applies to the pre-existing "already revoked" branch just above, which has the same problem.

Relatedly, `TestRotateRejectsExpiredCredential` asserts `StatusCode >= 400`, so a 500 passes and the wrong status class is never caught. Once the handler maps it, please tighten the assertion to the specific 4xx.

migrations/037_issued_credentials_audit_retention.up.sql
```diff
@@ -0,0 +1,29 @@
...
+-- pruned under the legacy expires_at rule until they age out. The backfill
+-- below stamps existing rows so history written before this migration
+-- survives too. 400 days mirrors the signing_credentials default
+-- (signing_credentials.audit_retention_days).
+
+ALTER TABLE issued_credentials
+    ADD COLUMN IF NOT EXISTS audit_retention_until TIMESTAMPTZ;
+
+UPDATE issued_credentials
+SET audit_retention_until = expires_at + INTERVAL '400 days'
```
saucam, line 46: The backfill hardcodes 400 days, but the runtime stamp uses the configured `token.audit_retention_days`. Any deployment running a non-default value gets a split: with 800d configured, all historical rows are pruned 400d earlier than policy; with 30d (data-minimization), historical rows are retained ~13x longer than the operator intends, with nothing flagging the mismatch.

The down.sql notes this for the down→up cycle, but the up-migration itself is the path every deployment takes. A migration can't read config, so at minimum add a comment here stating the fixed-400d assumption, plus an operator note (release notes or runbook) with a re-stamp UPDATE for deployments running a non-default window.

migrations/037_issued_credentials_audit_retention.up.sql
```diff
@@ -0,0 +1,29 @@
...
+WHERE audit_retention_until IS NULL;
+
+-- Retention pruning (delete only past the retention window).
+CREATE INDEX IF NOT EXISTS idx_issued_credentials_retention
```
saucam, line 50: Lock posture concern for large deployments: golang-migrate runs this file in one transaction, so the unbatched full-table `UPDATE` (every pre-037 row matches `IS NULL`) is followed by a plain `CREATE INDEX` whose SHARE lock blocks all INSERTs, i.e. token issuance, and the write-blocking window is backfill time + index build time combined. One dead tuple per row from the UPDATE also bloats the table until vacuum catches up.

`issued_credentials` is the per-token high-churn table, which is exactly the case migrations 017 and 033 call out as needing `CREATE INDEX CONCURRENTLY` outside the migration transaction (033's own comment: "if it ever needs to run online, switch to CREATE INDEX CONCURRENTLY"). At minimum, add the standard lock-posture comment those migrations carry; better, document a maintenance-window expectation or an out-of-band CONCURRENTLY path for big tables.

internal/service/credential.go
```diff
@@ -41,8 +46,15 @@ func NewCredentialService(
...
 	issuer string,
-	defaultTTL, maxTTL int,
+	defaultTTL, maxTTL, auditRetentionDays int,
 ) *CredentialService {
+	// A non-positive retention would stamp AuditRetentionUntil at or before
+	// expiry, silently degrading the graph back to delete-at-expiry. Clamp
+	// to the shipped default instead — misconfiguration must not erase
+	// evidence.
+	if auditRetentionDays <= 0 {
+		auditRetentionDays = 400
```
saucam: Two issues with this clamp.

First, the "0 = default 400" contract now lives in three unlinked layers: loadDefaults fills 400 when the key is absent, `Validate()` accepts 0 and documents the contract only in its error string, and this clamp is the sole place that actually enforces it. If the koanf default ever changes, this literal silently diverges and an explicit `audit_retention_days: 0` gets a different value than an absent key.

Second, an operator setting 0 for data-minimization reasons plausibly reads it as "no extra retention" and silently gets 400 days instead. That's a retention-policy footgun worth an explicit mention in the config docs, not just a code comment.

Suggest a named constant (the signing side already has `defaultSigningRetentionWindow`; this could be `defaultAuditRetentionDays` next to it) referenced from the clamp, and pointing the Validate() message and loadDefaults comment at it so there's one source of truth. The `INTERVAL '400 days'` in migration 037 can't share it, but a comment cross-referencing the constant would keep them honest.

internal/worker/cleanup.go
```diff
@@ -95,12 +98,20 @@ func (w *CleanupWorker) RunOnce(ctx context.Context) {
...
+	//   - audit_retention_until — the evidence clock: the delegation graph
+	//     (WalkUp/WalkDown/ListChains) reads expired rows within the audit
+	//     window, so historical chains stay answerable long after expiry.
+	// Rows written before migration 037 carry NULL audit_retention_until and
+	// fall back to the legacy expires_at-only rule. Backed by
+	// idx_issued_credentials_expires_at + idx_issued_credentials_retention.
 	credRes, err := w.db.NewDelete().
 		TableExpr("issued_credentials").
 		Where("expires_at < ?", now.Add(-w.credRetention)).
+		Where("(audit_retention_until IS NULL OR audit_retention_until < ?)", now).
```
saucam, line 114: Worth acknowledging the read-side blast radius of this predicate: for short-TTL tokens the effective retained population goes from ~90d (Token.MaxTTL cascade clock) to ~400d, roughly 4.4x. The delegation read path has no time bound at all — `ListByIdentity` has no expiry filter or LIMIT, `GetGraph`/`expandRootsByIdentity` seed `WalkDownMulti` with every parent-less JTI, and `MaxGraphNodes` trims only after the recursive CTE has done the work. An identity minting one root credential per session (the Claude Code pattern) will feel this directly in Delegation Explorer latency.

Not asking to change the retention model — it's the point of the PR — but a follow-up to put an `issued_at` window or LIMIT on the root-expansion path (or at least an issue tracking it) would keep the evidence window from becoming a graph-query regression.

One more nit while here: because the two clocks are ANDed, `audit_retention_days` below ~90 (MaxTTL) has no effect on deletion timing. Validate() accepts values down to 1, so maybe say in the config comment that the cascade clock is the floor.

saucam (commented):
Reviewed the two-clock model end to end. The core mechanism is sound and the AND-guard test matrix is good. A few things to address inline: the new expired-rotation rejection surfaces as a 500, the migration backfill/locking posture needs a look for large deployments, and the 400 default is now spelled out in several unlinked places.

KunalJavelin:
Thanks @saucam — this was a genuinely good review, all addressed in f4cfa2b:

1. **Expired/already-revoked rotation → 500.** Fixed. Added sentinels \`domain.ErrCredentialExpired\` (reused) and \`domain.ErrCredentialAlreadyRevoked\` (new); \`rotateCredentialOp\` now `errors.Is`-maps both to **409 Conflict** — a dead credential is a terminal-state client mistake, not a server fault. The pre-existing "already revoked" branch is fixed in the same move, as you flagged. `TestRotateRejectsExpiredCredential` now asserts `== 409`, not `>= 400`.

2. **Backfill hardcodes 400 vs configured window.** The migration can't read config, so 400 stays, but I added an explicit comment stating the fixed-400d assumption, the split it causes for non-default deployments (both directions), and an out-of-band re-stamp recipe + release-note pointer.

3. **Lock posture on 037.** Added the 017/033-style lock-posture comment (unbatched full-table UPDATE + non-CONCURRENT CREATE INDEX both write-block `issued_credentials` for their combined duration), documented the CONCURRENTLY/batched path for large installations, and added `SET LOCAL lock_timeout = '5s'` to fail fast rather than queue the auth plane. (Kept it a comment rather than restructuring, per your "at minimum" — the pre-037 cleanup bounds the table at the current fleet size, and migration-analyzer agreed.)

4. **"0 = 400" in three unlinked layers.** Fixed with one source of truth: `domain.DefaultAuditRetentionDays` (400) and `domain.MaxAuditRetentionDays` (36500), now referenced by `loadDefaults`, `Config.Validate` (message included), and the service clamp (which also gained the overflow ceiling Gemini flagged). On the data-minimization footgun: noted, but I kept 0→default rather than 0→"no retention" so a bare `audit_retention_days: 0` can never accidentally erase the evidence window — that seemed the safer default given the whole point of the change; happy to add a config-doc note if you disagree.

5. **Read-side blast radius + cascade-clock floor.** The floor nit is in code: `TokenConfig.AuditRetentionDays` doc now says the cleanup ANDs the two clocks so a value below MaxTTL (~90d) has no effect on deletion timing. The graph root-expansion bound is a real follow-up — tracked in #226.

Also fixed Gemini's real find: `token.audit_retention_days` was missing from the `loadEnvVars` integer-parse block, so `ZEROID_TOKEN_AUDIT_RETENTION_DAYS` was being stored as a string and bypassing fail-fast validation. Added (incidentally also covers `signing_credentials.audit_retention_days`, same suffix).

internal/service/credential.go
```diff
@@ -41,8 +46,15 @@ func NewCredentialService(
...
+	}
```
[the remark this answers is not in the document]
KunalJavelin: Fixed in f4cfa2b — NewCredentialService now clamps to domain.MaxAuditRetentionDays (36500) as well as the <=0→default rule, and Config.Validate shares the same bound, so the int64 Duration overflow is closed regardless of construction path.

config.go
```diff
@@ -634,6 +657,8 @@ func loadEnvVars(k *koanf.Koanf) error {
 		"ZEROID_ISSUER":                "token.issuer",
 		"ZEROID_TOKEN_TTL_SECONDS":     "token.default_ttl",
 		"ZEROID_MAX_TOKEN_TTL_SECONDS": "token.max_ttl",
+		// Evidence clock for issued_credentials (delegation-graph retention).
+		"ZEROID_TOKEN_AUDIT_RETENTION_DAYS": "token.audit_retention_days",
```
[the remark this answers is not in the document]
KunalJavelin: Fixed in f4cfa2b — .audit_retention_days added to the integer-parse block in loadEnvVars, so ZEROID_TOKEN_AUDIT_RETENTION_DAYS is parsed as an int and fails fast on bad input instead of falling through to a string.

migrations/037_issued_credentials_audit_retention.up.sql
```diff
@@ -0,0 +1,29 @@
...
+-- below stamps existing rows so history written before this migration
+-- survives too. 400 days mirrors the signing_credentials default
+-- (signing_credentials.audit_retention_days).
+
+ALTER TABLE issued_credentials
+    ADD COLUMN IF NOT EXISTS audit_retention_until TIMESTAMPTZ;
+
+UPDATE issued_credentials
+SET audit_retention_until = expires_at + INTERVAL '400 days'
+WHERE audit_retention_until IS NULL;
```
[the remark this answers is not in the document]
KunalJavelin: Addressed in f4cfa2b — added the 017/033-style lock-posture comment (the full-table backfill + index build hold write locks for their combined duration), SET LOCAL lock_timeout to fail fast, and documented the batched/out-of-band path for large installs. Bounded at the current fleet size by the pre-037 cleanup.

migrations/037_issued_credentials_audit_retention.up.sql
```diff
@@ -0,0 +1,29 @@
...
+
+-- Retention pruning (delete only past the retention window).
+CREATE INDEX IF NOT EXISTS idx_issued_credentials_retention
+    ON issued_credentials (audit_retention_until);
```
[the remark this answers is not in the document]
KunalJavelin: Addressed in f4cfa2b — the same lock-posture comment documents the CREATE INDEX CONCURRENTLY-in-its-own-migration path for large tables; kept non-concurrent for the current bounded table, with lock_timeout as the fast-fail guard.

internal/service/credential.go
KunalJavelin: Fixed in f4cfa2b — rotateCredentialOp now maps expired / already-revoked to 409 via sentinels domain.ErrCredentialExpired / ErrCredentialAlreadyRevoked, and the test asserts == 409. The pre-existing already-revoked branch is fixed in the same move. Good catch, thanks.

migrations/037_issued_credentials_audit_retention.up.sql
KunalJavelin: Addressed in f4cfa2b — the migration now documents the fixed-400d backfill vs the configured-window split, with an out-of-band re-stamp recipe + a release-note pointer for non-default deployments.

migrations/037_issued_credentials_audit_retention.up.sql
KunalJavelin: Addressed in f4cfa2b — 017/033-style lock-posture comment + SET LOCAL lock_timeout, and the CONCURRENTLY/batched path documented for large deployments.

internal/service/credential.go
KunalJavelin: Fixed in f4cfa2b — one source of truth now: domain.DefaultAuditRetentionDays (400) and MaxAuditRetentionDays (36500), referenced by loadDefaults, Config.Validate (error message included), and the service clamp. No more scattered 400 literal.

internal/worker/cleanup.go
KunalJavelin: Follow-up filed: zeroid#226 (bound the graph root-expansion read path with an issued_at window / LIMIT). The config doc now also notes the cascade clock (Token.MaxTTL) is the effective floor, per the nit.

saucam (approved):

Issue #138, opened 2026-05-13
security: SSRF guard on CIBA outbound HTTPS (block private-IP notification endpoints)


rsharath:
## Problem

CIBA ping/push outbound HTTPS POSTs (`dispatchPing` / `postCallback` in `internal/service/backchannel.go`) accept any HTTPS URL as `client_notification_endpoint` after `validateNotificationEndpoint` passes:

```go
// internal/service/oauth_client.go:287
func validateNotificationEndpoint(raw string) error {
    u, err := url.Parse(raw)
    if u.Scheme != "https" { return error }
    if u.Host == "" { return error }
    return nil
}
```

Acceptable URLs include:
- `https://10.0.0.5/` (RFC 1918 private IP)
- `https://localhost/`, `https://127.0.0.1/`
- `https://[::1]/`
- `https://169.254.169.254/` (AWS / GCP metadata server)
- `https://internal.corp/`

In **single-tenant** deployments this is acceptable: the endpoint is admin-registered, the operator knows the network topology, the risk is admin shooting themselves in the foot.

In **multi-tenant** deployments this is **SSRF**: an attacker who can register an OAuth client (or compromise a tenant admin) can use zeroid to scan the deployer's internal network or hit cloud-metadata services. The signed `client_notification_token` in the Authorization header makes the requests look legitimate to internal services.

### 1. Private-IP allowlist check

After URL parsing, resolve the host to an IP and reject if it falls in:
- IPv4 private ranges: `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `169.254.0.0/16`, `127.0.0.0/8`
- IPv4 reserved: `0.0.0.0/8`, `100.64.0.0/10` (CGNAT), `224.0.0.0/4` (multicast)
- IPv6 equivalents: `fc00::/7`, `fe80::/10`, `::1/128`, `::/128`

Implementation: `net.LookupIP(host)` → check each resolved IP against the rejection list. Reject the whole request if ANY resolved IP is private (DNS rebinding defense).

### 2. Make the guard opt-in via config

```yaml
backchannel:
  allow_private_notification_endpoints: false  # default; production
```

Single-tenant test/dev deployments can flip this to `true` to allow `https://localhost:3000/` style endpoints during integration testing.

### 3. Apply at BOTH registration and request-time

[...]

PR #228, opened 2026-07-09
feat: T5 — subagent scope-attenuation regression tests (monotonic requested ∩ parent)

KunalJavelin:
Part of highflame-ai/highflame-architecture#132 (epic), sub-task #138 (T5).

## What this PR lands
Integration regression tests in zeroid for **monotonic subagent scope attenuation** via `token_exchange` — the invariant that a delegated child can never out-scope its parent. (No `pkg/scopes` / personas package — that was intentionally dropped; the scope *vocabulary* lives in highflame-policy, and one credential policy covers every code agent.)

`tests/integration/subagent_delegation_test.go` asserts:
- **Attenuated child** — a child token requesting a subset gets exactly `requested ∩ parent`, with the act-chain deepened (`sub=subagent, act={sub=agent, act={sub=human}}`) and `depth+1`.
- **Over-broad request rejected** — a child requesting a scope the parent lacks is denied.
- **Depth cap** — delegation past the configured max depth is rejected.
- **Empty request grants nothing** — least-privilege is the structural default.

## Review follow-ups
- Closed response bodies + guarded nil responses in the test helpers (gemini review).
- Resolved the stale `pkg/scopes/*` threads — that package is not part of this PR.

## Tests
- [x] `GOEXPERIMENT=jsonv2 go test ./...` green (Postgres testcontainers); `go vet` clean.

### DIFF
--- a/tests/integration/subagent_delegation_test.go
+++ b/tests/integration/subagent_delegation_test.go
@@ -105,6 +105,8 @@ func TestSubagentDelegation_AttenuatedChildToken(t *testing.T) {
 	// by-jti walk: the child's edge must reference the parent (root) jti.
 	assert.Equal(t, rootJTI, parent["jti"], "sanity: root token jti must equal the returned rootJTI")
 	chainResp := get(t, adminPath("/delegations/by-jti/"+url.PathEscape(childJTI)), adminHeaders())
+	require.NotNil(t, chainResp)
+	defer func() { _ = chainResp.Body.Close() }()
 	require.Equal(t, http.StatusOK, chainResp.StatusCode)
 	chain := decode(t, chainResp)
 	edges, _ := chain["edges"].([]any)
@@ -145,6 +147,7 @@ func TestSubagentDelegation_OverBroadScopeRejected(t *testing.T) {
 		[]string{"tools:read", "tools:write"}, // child ceiling permits write…
 		[]string{"tools:write"},               // …but the parent never held it
 		rootTok)
+	require.NotNil(t, resp)
 	defer func() { _ = resp.Body.Close() }()
 
 	require.Equal(t, http.StatusBadRequest, resp.StatusCode,
@@ -171,6 +174,7 @@ func TestSubagentDelegation_EmptyRequestGrantsNothing(t *testing.T) {
 		[]string{"tools:read"}, // child ceiling permits read…
 		[]string{},             // …but the subagent requests no scopes at all
 		rootTok)
+	require.NotNil(t, resp)
 	defer func() { _ = resp.Body.Close() }()
 
 	require.Equal(t, http.StatusBadRequest, resp.StatusCode,
@@ -201,6 +205,7 @@ func TestSubagentDelegation_DepthCapRejected(t *testing.T) {
 	// Depth 2 exceeds the child policy's max_delegation_depth (1) — rejected.
 	resp := attemptSubagentExchange(t, capPolicy, "s9-depth-b",
 		[]string{"tools:read"}, []string{"tools:read"}, tokA)
+	require.NotNil(t, resp)
 	defer func() { _ = resp.Body.Close() }()
 
 	require.Equal(t, http.StatusBadRequest, resp.StatusCode,
--- a/tests/integration/subagent_delegation_test.go
+++ b/tests/integration/subagent_delegation_test.go
@@ -1,0 +1,211 @@
+package integration_test
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// These tests pin the subagent-delegation half of epic slice S9 (T5,
+// highflame-ai/highflame-architecture#138): a parent code-agent's delegated
+// token plus a child subagent's signed assertion are exchanged via
+// token_exchange (RFC 8693) for an ATTENUATED child token. They assert the
+// CAP-IDN-018 attenuation invariants specifically for the coding-agent scope
+// vocabulary (tools:*), on top of the generic delegation coverage in
+// delegation_test.go:
+//
+//   - child scopes = requested ∩ parent.granted ∩ child-policy.allowed
+//     (monotonic — a child can never hold a scope its parent lacks),
+//   - delegation_depth = parent.depth + 1, capped by the child credential
+//     policy's max_delegation_depth,
+//   - parent_jti references the parent credential, mission_id is inherited,
+//   - the act chain records the delegating parent (act.sub = parent.sub).
+//
+// The attenuation logic lives in OAuthService.tokenExchange +
+// CredentialService.IssueCredential + CredentialPolicyService.EnforcePolicy;
+// these tests are black-box regression coverage that it holds end-to-end for
+// the code-agent case. Tool-scope strings are used as literals here (as in
+// id_jag_test.go). The canonical tool-scope vocabulary lives in
+// highflame-policy (codegen'd), consumed by Shield/Guardian.
+
+// attemptSubagentExchange registers a fresh child subagent identity bound to
+// policyID, then performs a token_exchange WITHOUT asserting success — so
+// rejection-path tests can inspect the OAuth error. Mirrors the exchangeToken
+// helper in delegation_test.go minus the 200 assertion.
+func attemptSubagentExchange(t *testing.T, policyID, namePrefix string, childAllowedScopes, requestedScopes []string, parentToken string) *http.Response {
+	t.Helper()
+	key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	require.NoError(t, err)
+	extID := uid(namePrefix)
+	registerIdentityWithPolicy(t, extID, policyID, ecPublicKeyPEM(t, key), childAllowedScopes, adminHeaders())
+	wimse := fetchIdentityWIMSEByExternalID(t, extID)
+	return post(t, "/oauth2/token", map[string]any{
+		"grant_type":    "urn:ietf:params:oauth:grant-type:token-exchange",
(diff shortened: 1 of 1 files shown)


pkg/scopes/doc.go
```diff
@@ -0,0 +1,23 @@
+// Package scopes is the single source of truth for Highflame's coding-agent
+// authorization vocabulary: the canonical tools:* scope set and the persona
+// templates that map onto ZeroID credential policies.
+//
+// It is deliberately a standalone, zero-dependency module (like pkg/dpop and
+// pkg/authjwt) so downstream services can import it without pulling in
+// zeroid's server dependencies:
+//
+//   - Shield: TODO — Shield currently defines its own local tools:* consts
```
saucam: Two factual issues with this TODO:

1. Shield doesn't define local `tools:*` consts today, and there's no such TODO in shield — what exists is the `mcp:read`/`mcp:execute` action→scope map in `internal/scheduler/nhi.go`, which ADR 0005 D5 says gets *replaced* by `tools:*` and made catalog-driven. Worth correcting here and in the PR description so the remaining Shield migration work isn't understated.
2. "once zeroid publishes a pkg/scopes release tag" — this PR ships no release machinery for the new module (`release.yml` tag-submodules only tags pkg/authjwt; RELEASING.md untouched), so as merged this TODO is unreachable. See my review summary.

pkg/scopes/personas.go
```diff
@@ -0,0 +1,97 @@
...
+//	Persona.MaxDelegationDepth → credential policy max_delegation_depth
+//
+// Personas are the vocabulary used when CREATING those policies — e.g. when
+// Guardian/Cerberus selects a persona at subagent registration, or when an
+// operator provisions a code-agent identity. Runtime attenuation in
+// token_exchange reads the resulting credential policy, never the persona, so
+// this package carries no enforcement logic — only the canonical templates
+// that Shield (scope ceiling) and Guardian (persona selection) import as one
+// source of truth.
+type Persona struct {
```
saucam: This struct is a strict subset of credential policy's fields (`allowed_scopes` + `max_delegation_depth`), and per the comment above, enforcement never reads it — only the credential policy created from it. That makes the compiled table an advisory copy that can silently diverge from the enforced policy (operator edits the `code-editor` policy in the DB; every service holding this table still displays the old authority).

Alternative that needs no new concept: seed these four as well-known credential policies, keep only a `DefaultPersona` name constant + fail-closed lookup for unmapped subagents. Also solves tenant extensibility (a custom persona is just another policy) and avoids the release cascade of a persona tweak needing a zeroid tag + dep bumps in every consumer. Expanded in my review summary + the ADR comment on arch#219.

pkg/scopes/scopes_test.go
```diff
@@ -0,0 +1,67 @@
...
+		if !IsValidToolScope(s) {
+			t.Errorf("IsValidToolScope(%q) = false, want true", s)
+		}
+	}
+
+	invalid := []string{
+		"",
+		"tools",
+		"tools:",
+		"tools:exec",  // near-miss: canonical scope is tools:execute
```
saucam: This near-miss is live in this repo: `tests/integration/id_jag_test.go` uses `tools:exec` in 4 places (lines 102, 132, 436, 454) on main. Since nothing validates scope strings at mint, both spellings keep "working" while never matching each other — a credential minted with `tools:exec` will silently never match a ceiling built on `tools:execute`. If `tools:execute` is canonical, this PR should migrate `id_jag_test.go` in the same change.

tests/integration/subagent_delegation_test.go
```diff
@@ -0,0 +1,185 @@
...
+//   - delegation_depth = parent.depth + 1, capped by the child credential
+//     policy's max_delegation_depth,
+//   - parent_jti references the parent credential, mission_id is inherited,
+//   - the act chain records the delegating parent (act.sub = parent.sub).
+//
+// The attenuation logic lives in OAuthService.tokenExchange +
+// CredentialService.IssueCredential + CredentialPolicyService.EnforcePolicy;
+// these tests are black-box regression coverage that it holds end-to-end for
+// the code-agent case. Tool-scope strings are used as literals here (as in
+// id_jag_test.go) to keep the main module decoupled from pkg/scopes — the
```
saucam: The decoupling rationale doesn't hold for test-only imports — RELEASING.md and release.yml document the exact precedent with pkg/authjwt: "imported only from tests/integration/. Go doesn't follow test imports across module boundaries, so its version reference in zeroid/go.mod is invisible to downstream consumers." Importing pkg/scopes here would cost consumers nothing and would make these tests actually pin the canonical consts — the literals are what allow the `tools:exec`/`tools:execute` drift that already exists in id_jag_test.go.

tests/integration/subagent_delegation_test.go
```diff
@@ -0,0 +1,185 @@
...
+// canonical vocabulary lives there for Shield/Guardian to import.
+
+// attemptSubagentExchange registers a fresh child subagent identity bound to
+// policyID, then performs a token_exchange WITHOUT asserting success — so
+// rejection-path tests can inspect the OAuth error. Mirrors the exchangeToken
+// helper in delegation_test.go minus the 200 assertion.
```
saucam, line 40: Rather than a near-clone, consider inverting: make `delegation_test.go`'s `exchangeToken` wrap this helper (`resp := attemptSubagentExchange(...)`; assert 200; decode), so the register→WIMSE→assertion→POST sequence exists once and the next change to the exchange request shape is a single edit.

saucam (commented):
**Part A (the attenuation tests) looks great.** I cross-checked the invariants and error codes against main and ran the suite locally: all three `TestSubagentDelegation_*` tests pass end-to-end and `invalid_scope` / `policy_violation` / 400 match what the server actually returns. Happy for that half to land as-is.

My concerns are with Part B and CI:

**1. CI never runs the new module's tests.** In workspace mode, `go list ./...` from the repo root returns only root-module packages; workspace member modules are excluded. Verified with the CI-identical env (go 1.26, `GOEXPERIMENT=jsonv2`): the unit-test step in `pr-check.yml` (`go test $(go list ./... | grep -v '/tests')`) yields zero `pkg/*` packages, so `personas_test.go` and `scopes_test.go` are silently skipped. The step's own comment claims it covers pkg/dpop and pkg/authjwt unit tests, but the hole pre-exists for those two as well. Since these tests are the pinning mechanism for the catalog, they need to actually execute — please add explicit per-module test steps (`go test ./pkg/scopes/...`, and ideally fix dpop/authjwt in the same breath).

**2. RELEASING.md checklist is skipped.** RELEASING.md opens with "ZeroID is published as **three Go modules**" and has an explicit checklist for adding a nested module; this PR does only the go.work step. `release.yml`'s tag-submodules step tags only pkg/authjwt, so as merged there is no path to the `pkg/scopes` release tag that doc.go's TODO depends on, and `go get .../pkg/scopes@latest` resolves no versions. If the module stays, it needs the checklist completed (RELEASING.md table + tag-submodules or a release-scopes.yml).

**3. The bigger question — should this module live in zeroid at all?** Raising here and on the ADR (see my comment on highflame-ai/highflame-architecture#219). Two parts:

- *Scope catalog placement.* zeroid's token path is deliberately vocabulary-agnostic (requested ∩ granted ∩ policy set-math over opaque strings) — that is the correct AS design, and nothing in zeroid, including this PR's own integration tests, imports pkg/scopes. The actual consumers are all platform-side and multi-language: Shield's ceiling (Go), Guardian/ramparts (Rust), Cerberus, Studio's persona picker (TypeScript). A Go-only module can't be the source of truth for the TS and Rust consumers — they'll re-type the strings, which is exactly the drift this package exists to prevent (and which already exists in-repo, see the `tools:exec` inline comment). highflame-policy already codegens catalogs to Rust/TS/Go/Py; that looks like the right home for the vocabulary, with zeroid staying vocabulary-agnostic.

- *Personas.* A `Persona` is `{name, allowed_scopes, max_delegation_depth}` — a strict subset of credential policy's fields, and the doc comment itself notes runtime enforcement "reads the resulting credential policy, never the persona". So the compiled table is an advisory copy that diverges the moment an operator edits the `code-editor` policy in the DB. Same outcome with no new concept: seed the four personas as well-known credential policies (tag them for the Studio picker), keep a `DefaultPersona` *name* constant with fail-closed lookup for unmapped subagents. Tenants then get custom personas for free (it's just another policy row), and there's no second authority to keep in sync.

None of this blocks Part A — I'd be glad to see the tests split out and landed independently while we settle Part B.

tests/integration/subagent_delegation_test.go
KunalJavelin: attemptSubagentExchange is intentionally retained as a near-duplicate rather than having exchangeToken wrap it. Keeping the subagent regression helper independent of delegation_test.go`s broadly-used exchangeToken isolates these tests: a future change to the exchange request shape here cannot ripple into the many exchangeToken callers, and vice versa. Retaining the small duplication for test isolation.

tests/integration/subagent_delegation_test.go
KunalJavelin: pkg/scopes was intentionally dropped from this PR (personas removed; the tools:* scope vocabulary now lives in highflame-policy, codegen`d and consumed by Shield/Guardian, not in zeroid). With no pkg/scopes module to import, the hardcoded tools:* literals are the correct form here. Resolving as moot.

tests/integration/subagent_delegation_test.go
```diff
@@ -0,0 +1,185 @@
+package integration_test
```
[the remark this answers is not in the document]
KunalJavelin: pkg/scopes was intentionally removed from this PR — the scope vocabulary now lives in highflame-policy (codegen`d, consumed by Shield/Guardian), not in zeroid. There is no pkg/scopes module for these tests to import, so the hardcoded tools:* string literals are the correct and intended form (matching id_jag_test.go). Resolving.

saucam (approved):

PR #241, opened 2026-07-21
feat: Client ID Metadata Documents (CIMD) — zero-registration client onboarding

rsharath:
## Summary

Implements **Client ID Metadata Documents** ([draft-ietf-oauth-client-id-metadata-document](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/)), the client-onboarding model the MCP Authorization spec (2025-11-25) names as its preferred default.

A client completes an `authorization_code` + PKCE flow against ZeroID **with zero pre-registration**: it uses a stable `https://` URL as its `client_id`; ZeroID fetches the JSON metadata document published at that URL, validates it, and treats it as an ephemeral public PKCE client. Nothing is persisted — the trust model is domain ownership via TLS + DNS, with the document's `redirect_uris` as the primary anti-impersonation control.

Full reference: [`docs/cimd.md`](docs/cimd.md).

## How it works

- **Registry-first, CIMD fallback.** CIMD resolution only runs when the client registry misses AND the `client_id` is CIMD-shaped (absolute `https://` URL with a path). Existing URL-shaped registry clients keep working; deployers can pin/override a CIMD URL by registering it.
- **SSRF-guarded fetch** reusing the audited DNS-rebinding-safe client from attestation (`attestation.NewSSRFGuardedHTTPClient`): private/loopback/link-local/metadata ranges blocked, connection pinned to the validated IP, 5 KiB size cap, 5 s timeout, **redirects refused**, 255-char `client_id` cap (protects `VARCHAR(255)` columns downstream).
- **Document validation:** self-reference (`client_id` field must equal the fetch URL), `redirect_uris` required + OAuth 2.1 scheme rules (https / loopback http / private-use scheme), public PKCE only (`token_endpoint_auth_method: none`), grant allow-list `authorization_code`/`refresh_token`, `response_types` must include `code`.
- **Bounded memoization:** 1000-entry in-memory cache; success TTL 1 h (24 h hard cap), 60 s negative caching so replaying a dead URL can't force repeated timeout-bounded outbound fetches.
- **Full lifecycle:** `refresh_token` grant re-resolves the document before rotation — unpublishing the document ends rotation (without consuming the presented token on transient failures). Introspection/revocation accept a CIMD `client_id` as public-client identification (RFC 7662/7009 §2.1).

## Configuration

On by default (`cimd.enabled: true`); disable with `ZEROID_CIMD_ENABLED=false`. Optional `allowed_domains` allowlist for closed ecosystems; test/dev-only `allow_private_metadata_endpoints` SSRF relaxation; size/TTL knobs. Advertised via `client_id_metadata_document_supported` in `/.well-known/oauth-authorization-server` when enabled.

## Testing

- Unit tests cover detection, the full validation matrix (self-reference mismatch, missing/invalid `redirect_uris`, confidential auth methods, disallowed grants), size cap, non-200, redirect refusal, negative caching, cache expiry + eviction at cap, domain allowlist, and TTL clamps.
- Full suite passes with `-race`; `go vet` + `gofmt` clean.


jalbrethsen-highflame:
## Deactivated registry clients are resurrected by the CIMD fallback

**TL;DR:** the new registry-miss → CIMD fallback in `IssueAuthCode` and `VerifyPresentedClientAuth` can't tell "client was never registered" apart from "client was deactivated" — so deactivating a registry client whose `client_id` is a CIMD-shaped URL doesn't kill it. The next `/oauth2/authorize` just fetches its metadata document and re-synthesizes it as an ephemeral CIMD client.

### What happens

`GetPublicClient` (untouched by this PR) collapses two distinct cases into one error:

```go
// internal/service/oauth_client.go
client, err := s.repo.GetPublicByClientID(ctx, clientID)
if err != nil {
    return nil, ErrOAuthClientNotFound   // row doesn't exist
}
if !client.IsActive {
    return nil, ErrOAuthClientNotFound   // row exists, deployer deactivated it
}
```

Before this PR that conflation was harmless — both cases ended at the same 401. The new fallback in `IssueAuthCode` makes the distinction load-bearing for the first time:

```go
oauthClient, err := s.oauthClientSvc.GetPublicClient(ctx, req.ClientID)
if err != nil {
    if !s.cimdSvc.Enabled() || !IsCIMDClientID(req.ClientID) {
        return "", oauthUnauthorized("unknown or inactive client_id", err)
    }
    oauthClient, err = s.cimdSvc.ResolveClient(ctx, req.ClientID)   // ← also runs for deactivated clients
    ...
}
```

Consequences:

1. **Deactivation stops being a kill switch.** Soft-delete (`is_active = false`) is the platform convention for revoking a client, and it's exactly the moment a deployer is trying to shut one down. For a URL-`client_id` client, the flow now falls through to the document and carries on.
2. **It breaks the PR's own pinning story.** `docs/cimd.md` documents "registering a client under a CIMD URL overrides the document." Deactivating that pin silently hands control back to the document — the opposite of what the deployer intended.
3. **The two halves of the flow disagree.** `authorizationCode` (the token path) gets this right: `GetClientByClientID` returns inactive rows with `err == nil`, the new `switch` checks `IsActive`, and refuses **without** CIMD fallback. So an inactive URL-client can be issued an auth code but can't exchange it — the authorize path is the wrong half.

Same pattern in the `VerifyPresentedClientAuth` arm (introspection/revocation): `GetPublicClient` → `ErrOAuthClientNotFound` → CIMD resolve → auth passes for a deactivated client.

### Suggested fix

Only fall back to CIMD when the registry truly has **no row** — which is exactly the check the token path already does. In the fallback branch, consult `GetClientByClientID` (returns inactive rows with `err == nil`) before resolving:

```go
oauthClient, err := s.oauthClientSvc.GetPublicClient(ctx, req.ClientID)
if err != nil {
    // Fall back to CIMD only when the registry has no row at all. A row that
    // exists but is inactive (or non-public) was deliberately registered —
    // deactivation must remain a kill switch, and a registered row must keep
    // its documented precedence over the metadata document.
    if !s.cimdSvc.Enabled() || !IsCIMDClientID(req.ClientID) {
        return "", oauthUnauthorized("unknown or inactive client_id", err)
    }
    if _, rowErr := s.oauthClientSvc.GetClientByClientID(ctx, req.ClientID); rowErr == nil {
        return "", oauthUnauthorized("unknown or inactive client_id", err)
    }
    oauthClient, err = s.cimdSvc.ResolveClient(ctx, req.ClientID)
    if err != nil {
        return "", cimdOAuthError(err)
    }
}
```

(Alternatively, split the sentinel — e.g. `ErrOAuthClientInactive` vs `ErrOAuthClientNotFound` — and gate the fallback on `errors.Is(err, ErrOAuthClientNotFound)`. That's cleaner long-term but touches `oauth_client.go`'s contract; the row-existence check keeps the fix inside this PR's footprint.)

Apply the same guard to the `VerifyPresentedClientAuth` CIMD arm. Worth a regression test either way: register a client with a CIMD-URL `client_id`, deactivate it, and assert both `IssueAuthCode` and `VerifyPresentedClientAuth` return 401 without any outbound document fetch (the httptest hit counter in `cimd_test.go` is perfect for asserting "no fetch happened").

internal/service/cimd.go
```diff
@@ -0,0 +1,525 @@
...
+			return nil
+		}
+		return fmt.Errorf("redirect_uri %q: http is only permitted for loopback hosts", raw)
+	case "":
+		return fmt.Errorf("redirect_uri %q must be an absolute URI", raw)
+	default:
+		// Private-use scheme (RFC 8252 §7.1) — native app callback.
+		return nil
+	}
+}
```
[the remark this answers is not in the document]
rsharath: Fixed in 196a4b5. `validateCIMDRedirectURI` now, for the `http`/`https` cases, requires an absolute URI with a host and rejects userinfo, and rejects a fragment regardless of scheme. `url.Parse("https:/cb")` yields `Host == ""`, so the rootless form is now rejected; `https://user:pw@host/cb` (userinfo) and `https://host/cb#frag` (fragment) are rejected too. Added those cases to `TestValidateCIMDRedirectURI`.

internal/service/cimd.go
```diff
@@ -0,0 +1,525 @@
...
+	}
+	if s.now().After(e.expiresAt) {
+		delete(s.cache, clientID)
+		return nil, nil, false
+	}
+	if e.err != nil {
+		return nil, e.err, true
+	}
+	cp := *e.client
+	return &cp, nil, true
```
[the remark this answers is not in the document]
rsharath: Fixed in 196a4b5 (covers both this site and line 519/522). Added `cloneCIMDClient`, which does the struct copy and then `slices.Clone`s the slice fields (`GrantTypes`/`RedirectURIs`/`Scopes`/`Contacts`), and use it on both `storeResult` (so the cache doesn't alias the synthesized client passed in) and `cachedResult` (so each returned copy is independent). New `TestCIMDCacheReturnsIndependentCopies` mutates every slice on a returned client and asserts the next cache hit is unaffected.

internal/service/oauth.go
```diff
@@ -1461,12 +1501,26 @@ func (s *OAuthService) IssueAuthCode(ctx context.Context, req IssueAuthCodeReque
...
+	// working unchanged (CIMD is on by default), and gives deployers a pinning
+	// mechanism: registering a client under a CIMD URL overrides the document.
+	// CIMD clients are inherently public PKCE clients, so both paths converge
+	// on the same shape.
 	oauthClient, err := s.oauthClientSvc.GetPublicClient(ctx, req.ClientID)
 	if err != nil {
-		return "", oauthUnauthorized("unknown or inactive client_id", err)
+		if !s.cimdSvc.Enabled() || !IsCIMDClientID(req.ClientID) {
+			return "", oauthUnauthorized("unknown or inactive client_id", err)
+		}
```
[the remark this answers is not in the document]
rsharath: Good catch — fixed in 1446a28 (same root cause as the deactivation report above). `IssueAuthCode` now falls back to CIMD only when the registry has **no row at all**: after `GetPublicClient` misses, it consults `GetClientByClientID` (which returns any row — public or confidential, active or not — with `err == nil`) and refuses if a row exists. So a confidential registry client whose `client_id` happens to be an https:// URL keeps its precedence and can't be shadowed by a CIMD document, eliminating the mid-flow authorize/token mismatch. Regression test asserts the authorize path 401s with zero outbound document fetch.

internal/service/oauth.go
```diff
@@ -2269,15 +2347,21 @@ func (s *OAuthService) VerifyPresentedClientAuth(ctx context.Context, clientID,
...
+	// confidential client skip authentication.
 	if clientSecret == "" {
 		if _, err := s.oauthClientSvc.GetPublicClient(ctx, clientID); err != nil {
 			if errors.Is(err, ErrOAuthClientNotFound) {
+				if s.cimdSvc.Enabled() && IsCIMDClientID(clientID) {
+					if _, cimdErr := s.cimdSvc.ResolveClient(ctx, clientID); cimdErr == nil {
+						return nil
+					}
+				}
 				return oauthUnauthorized("client authentication required", nil)
```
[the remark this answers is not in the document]
rsharath: Fixed in 1446a28 — the same row-existence guard is applied to the `VerifyPresentedClientAuth` CIMD arm. A no-secret caller can now only be treated as a CIMD public client when the registry has no record for that `client_id`; if a confidential client is registered under a CIMD-shaped URL, `GetClientByClientID` finds it and the caller must present its `client_secret`, closing the introspection/revocation auth-bypass. Covered by the regression test (introspection path 401s with zero document fetch).

rsharath:
## Review responses — all comments addressed (2 commits)

Thanks @jalbrethsen-highflame and the Copilot reviewer. Pushed two commits; per-thread replies are inline above.

### 1446a28 — deactivation kill-switch (fixes @jalbrethsen-highflame's report + Copilot's two `oauth.go` shadowing threads)

Root cause exactly as described: `GetPublicClient` collapses "no row" and "row exists but inactive/non-public" into one `ErrOAuthClientNotFound`, and the new CIMD fallback treated that as "no row". Applied the row-existence guard you suggested — fall back to CIMD only when `GetClientByClientID` confirms there is genuinely no row — to **both** `IssueAuthCode` and `VerifyPresentedClientAuth`. This restores deactivation as a kill switch, preserves the documented pinning precedence, and closes the confidential-client shadowing / introspection-bypass Copilot flagged (same root cause).

Regression test `TestCIMD_DeactivatedRegistryClientNotResurrected` (integration, testcontainers): register a client under a CIMD-URL `client_id` → deactivate → assert both paths 401 **with zero outbound document fetch** (httptest hit counter), plus a positive control that an unregistered CIMD URL still resolves. Verified it fails on the pre-fix code ("An error is expected but got nil") and passes after.

### 196a4b5 — remaining Copilot findings

- **`validateCIMDRedirectURI`** — now requires an absolute URI with a host for http/https and rejects fragments + userinfo (`https:/cb`, `https://user:pw@host/cb`, `…/cb#frag` all rejected). Cases added to `TestValidateCIMDRedirectURI`.
- **Cache slice aliasing** (`cachedResult`/`storeResult`) — added `cloneCIMDClient` (deep-copies `GrantTypes`/`RedirectURIs`/`Scopes`/`Contacts`), used on both store and read. New `TestCIMDCacheReturnsIndependentCopies`.
- **AS discovery metadata** (from the review-body notes) — `response_types_supported` was `["token"]` though `/oauth2/authorize` requires `code`; `grant_types_supported` omitted `authorization_code`/`refresh_token`; `token_endpoint_auth_methods_supported` omitted `none`. All three corrected — these directly affect the discovery-driven MCP clients CIMD targets. RFC 8414 compliance suite still green.

`go vet` + `gofmt` clean; service unit tests and the RFC 8414 / CIMD integration tests pass locally (`GOEXPERIMENT=jsonv2`).

internal/service/cimd.go
```diff
@@ -0,0 +1,549 @@
...
+	_, ok := s.allowedDomains[strings.ToLower(host)]
+	return ok
+}
+
+// fetch performs the SSRF-guarded GET and JSON-decodes the (size-capped) body.
+func (s *CIMDService) fetch(ctx context.Context, docURL string) (*cimdMetadataDocument, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, docURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrCIMDFetch, err)
+	}
```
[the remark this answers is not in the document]
rsharath: Fixed in d44b068. `fetch` now wraps the context with `context.WithTimeout(ctx, cimdFetchTimeout)` (5s), so the documented bound holds regardless of the injected client's own `Timeout` (the production SSRF-guarded client is 10s) — effective deadline is `min(5s, client.Timeout)`, and a slow origin can't stall the request path past 5s.

internal/handler/wellknown.go
```diff
@@ -140,9 +145,11 @@ func (a *API) oauthMetadataOp(_ context.Context, _ *struct{}) (*OAuthMetadataOut
 		// RFC 8414 — client auth methods accepted by the introspection and
 		// revocation endpoints (client_secret_basic via Authorization header,
 		// client_secret_post via body fields).
-		"introspection_endpoint_auth_methods_supported":    []string{"client_secret_post", "client_secret_basic"},
-		"revocation_endpoint_auth_methods_supported":       []string{"client_secret_post", "client_secret_basic"},
-		"response_types_supported":                         []string{"token"},
+		"introspection_endpoint_auth_methods_supported": []string{"client_secret_post", "client_secret_basic"},
+		"revocation_endpoint_auth_methods_supported":    []string{"client_secret_post", "client_secret_basic"},
```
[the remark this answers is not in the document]
rsharath: Fixed in d44b068 — added `"none"` to both `introspection_endpoint_auth_methods_supported` and `revocation_endpoint_auth_methods_supported`, matching that `VerifyPresentedClientAuth` accepts a no-secret public/CIMD `client_id` on those endpoints (RFC 7009 §2.1 / RFC 7662 §2.1). Consistent with the `none` we added to `token_endpoint_auth_methods_supported`.

rsharath:
## Second review pass addressed — d44b068

Thanks Copilot — the two suppressed notes on `oauth.go:1530` / `oauth.go:2379` are a real hole in the first fix, and it's exactly the sentinel-split @jalbrethsen-highflame flagged as the cleaner long-term option. Addressed properly:

**DB-outage error conflation (the `GetClientByClientID` gate).** `OAuthClientService.GetClientByClientID` mapped *every* repo error to `ErrOAuthClientNotFound`, so under a store failure my `rowErr`-based guard read "no row" and could still resurrect a deactivated CIMD-shaped client. It now returns `ErrOAuthClientNotFound` only for `sql.ErrNoRows` and surfaces operational errors as-is. All three CIMD-decision sites now fall back to CIMD **only** on `errors.Is(err, ErrOAuthClientNotFound)` and fail closed otherwise:
- `IssueAuthCode` — operational error → 500, not CIMD.
- `authorization_code` token path — same (this had the same latent bug via the old `switch`, now gated).
- `VerifyPresentedClientAuth` — operational error → reject, not CIMD.

The `refresh_token` path already gated on `errors.Is(...ErrOAuthClientNotFound)`, and the three non-CIMD callers of `GetClientByClientID` (`EnsureClient`, two CIBA sites) only test `err != nil`, so the contract change is behaviour-preserving for them.

**Also in d44b068** (the two inline threads above): CIMD `fetch` now enforces the documented 5s bound via `context.WithTimeout` regardless of the injected client's `Timeout`; and `introspection_/revocation_endpoint_auth_methods_supported` now include `"none"`.

Build + `go vet` + `gofmt` clean; CIMD unit tests, the deactivation regression test, and the RFC 8414 / CIBA discovery compliance suite all pass locally (`GOEXPERIMENT=jsonv2`, testcontainers Postgres).

saucam (approved):

internal/service/cimd.go
```diff
@@ -0,0 +1,605 @@
...
+	for _, ru := range doc.RedirectURIs {
+		if err := validateCIMDRedirectURI(ru); err != nil {
+			return nil, fmt.Errorf("%w: %v", ErrCIMDInvalidDocument, err)
+		}
+	}
+
+	// Public client only in v1. An omitted token_endpoint_auth_method defaults to
+	// "none" for a CIMD (public) client; any explicit confidential method is
+	// rejected — ZeroID does not accept private_key_jwt token-endpoint auth yet.
```
adeinega, line 446: The OAuth Client ID Metadata Document spec (draft 2) mentions only `token_endpoint_auth_method` and for an unknown reason doesn't `revocation_endpoint_auth_methods_supported` and `introspection_endpoint_auth_methods_supported` (per RFC 8414) so that `private_key_jwt`, as a client auth method, could be used for revocation and introspect endpoints as well.

docs/cimd.md
```diff
@@ -0,0 +1,178 @@
...
+```http
+POST /oauth2/authorize HTTP/1.1
+Content-Type: application/x-www-form-urlencoded
+
+client_id=https%3A%2F%2Fapp.example.com%2Foauth%2Fclient.json
+&redirect_uri=http%3A%2F%2F127.0.0.1%3A3000%2Fcallback
+&response_type=code
+&code_challenge=<S256 challenge>
+&code_challenge_method=S256
+&state=<opaque>
```
adeinega, line 57: No need to pass `state` here...

internal/handler/wellknown.go
```diff
@@ -139,10 +144,15 @@ func (a *API) oauthMetadataOp(_ context.Context, _ *struct{}) (*OAuthMetadataOut
...
-		"revocation_endpoint_auth_methods_supported":       []string{"client_secret_post", "client_secret_basic"},
-		"response_types_supported":                         []string{"token"},
+		// client_secret_post via body fields). "none" is included because
+		// VerifyPresentedClientAuth accepts a no-secret REGISTERED public
+		// client_id on these endpoints (RFC 7009 §2.1 / RFC 7662 §2.1).
+		// CIMD client_ids are not accepted there — see VerifyPresentedClientAuth.
+		"introspection_endpoint_auth_methods_supported": []string{"client_secret_post", "client_secret_basic", "none"},
+		"revocation_endpoint_auth_methods_supported":    []string{"client_secret_post", "client_secret_basic", "none"},
+		// /oauth2/authorize enforces response_type=code (OAuth 2.1 — the
+		// implicit "token" flow is not supported), so advertise "code".
```
adeinega, line 154: On a side note, while OAuth 2.1 dropped it, the flow remains to be there in OpenID Connect 1.0, and I doubt it will ever go away anytime soon.

I see nothing wrong with passing encrypted ID Tokens via front channel.

internal/service/cimd.go
```diff
@@ -0,0 +1,605 @@
...
+	}, nil
+}
+
+// validateCIMDRedirectURI enforces OAuth 2.1 §2.3-style redirect URI scheme
+// rules on a CIMD document entry: https:// always OK; http:// only for
+// loopback hosts (native/CLI callbacks per RFC 8252 §7.3); any other non-empty
+// scheme is treated as a private-use scheme (native apps, e.g. myapp:/cb).
+// Relative URIs are rejected — a redirect target must be absolute.
+func validateCIMDRedirectURI(raw string) error {
+	u, err := url.Parse(raw)
```
adeinega, line 511: `ParseRequestURI` can save a few lines of code, and omit `if u.Fragment != "` and a few other things.

adeinega:
There is a lot to unpack... to start off, `AllowedDomains` in cimd.go (and the conf) is nothing but a poor man's defense against malicious actors who can either reuse existing `client_id`, or somehow manage to run a simple HTTP service that returns a client's metadata within an org to be able to start and then use his public client (public clients use 127.0.0.1 in their redirect URLs).

I'd say that the spec [itself](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/) needs to put more thoughts in that. The whole thing with public / native clients looks too clumsy for me, if a user wants to run a native client (using CIMD) on his machine that it's a bit hard (but not impossible) to run a service that available using a proper cert + there is no guarantee that this machine will be reachable for an AS (ZeroID). If some other (internal or external) service makes available this metadata for all clients, we immediately lose a ton of benefits such as a client's JWKs, say for "private_key_jwt" as a client's auth method, or keys for encryption.

docs/cimd.md
rsharath: Right, and the doc now says so — fixed in #278 (`docs/cimd.md`).

`state` is optional here. PKCE is mandatory at this endpoint (`code_challenge` and `code_challenge_method` are both in the required-field gate; `state` is not), and PKCE carries the CSRF binding `state` was originally needed for — so OAuth 2.1 leaves `state` to carry application state only.

Kept in the example rather than removed, since most clients have somewhere to return the user to and ZeroID round-trips it verbatim, but it no longer reads as required.

internal/service/cimd.go
rsharath: Checked this one empirically, and it goes the wrong way — `ParseRequestURI` would make the fragment check vacuous rather than unnecessary.

It does not populate `Fragment`. Per its own doc, it assumes the input has no `#fragment` suffix, so the fragment is swallowed into `Path`:

```
input                             | url.Parse                | url.ParseRequestURI
https://app.example.com/cb#frag   | h="app.example.com"      | h="app.example.com"
                                  | f="frag" p="/cb"         | f="" p="/cb#frag"
```

So dropping `if u.Fragment != ""` and switching parsers would let `https://app.example.com/cb#frag` through validation with the fragment hidden in the path — and `TestValidateCIMDRedirectURI`s fragment case would pass while asserting nothing.

It also does not save the other checks. `https:/cb` still yields `Host == ""`, and `/relative/cb` is accepted outright (`ParseRequestURI` permits an absolute path), so the `Host == ""` guard is still what rejects both. Userinfo is unaffected either way.

Staying with `url.Parse`.

internal/handler/wellknown.go
rsharath: Agreed on the substance, and it does not apply to ZeroID today: it issues no ID Tokens at all. It is an OAuth 2.1 AS, not an OpenID Provider — there is no `id_token` code path, no `openid` scope handling, and no `id_token`-bearing response type. So the front-channel question, encrypted or not, never arises here.

Worth stating plainly because it is easy to assume otherwise from the metadata surface: for the interop work we can claim the OAuth AS and Resource AS columns, not the OpenID Provider one. If ZeroID ever grows an OP mode, this is the right thing to revisit — but the response-type set would be a deliberate decision then, not an inheritance from what we advertise now.

internal/service/cimd.go
rsharath: Good catch on the spec detail — CIMD draft 2 defines only `token_endpoint_auth_method`, and RFC 8414 does carry separate `revocation_endpoint_auth_methods_supported` / `introspection_endpoint_auth_methods_supported`, so the document has no way to say "public at the token endpoint, `private_key_jwt` at revocation".

Nothing to change while v1 stands, since the constraint is upstream of that: ZeroID accepts no `private_key_jwt` anywhere, so the code rejects any `token_endpoint_auth_method` other than `none` and every CIMD client is public. `VerifyPresentedClientAuth` correspondingly does not accept CIMD `client_id`s at introspection/revocation at all — the `none` in those metadata arrays is for REGISTERED public clients only, which the comment at `wellknown.go` now spells out.

Carried onto #264 (confidential CIMD clients: parse `jwks_uri`/`jwks`, accept `private_key_jwt`), because it changes what that work has to decide: whether one `token_endpoint_auth_method` in the document governs the other two endpoints, or whether we only honour it at the token endpoint and keep CIMD clients public elsewhere. Thanks — that would have been easy to miss until it was already built.

PR #244, opened 2026-07-23
feat: RFC 8707 resource indicators for OAuth 2.0 token endpoint

### LINKED ISSUE #199: OAuth access tokens: aud defaults to the issuer — adopt RFC 8707 resource indicators so it names the target resource server
## Problem

When the credential service issues a token without an explicit audience, `aud` falls back to the **issuer URL** — so by default `aud == iss`. That is a present-but-nominal placeholder, not a real audience. Per [RFC 7519 §4.1.3](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3), `aud` identifies the *recipients the token is intended for* (the target resource server), not the party that minted it (that is `iss`).

```go
// internal/service/credential.go:375-382
// JWT-SVID §3 requires `aud` to be present on every issued token. Default
// to the issuer URL when no audience was supplied so tokens remain
// interoperable with spec-compliant verifiers (e.g., pkg/authjwt).
aud := req.Audience
if len(aud) == 0 {
    aud = []string{s.issuer}
}
_ = token.Set(jwt.AudienceKey, aud)
```

The root cause is that the OAuth token endpoint gives a caller no way to name the target resource. `TokenInput.Body` (`internal/handler/oauth.go:26-65`) exposes no `resource` or `audience` parameter on any grant, including token-exchange (RFC 8693). So at mint time the AS does not know the intended RS, and the JWT-SVID §3 "`aud` MUST be present" rule forces a non-empty fallback — the issuer.

## Impact

- **Semantically wrong:** the token effectively says "intended for the issuer," which is never the real relying party.
- **Latent rejection risk:** a strict resource server that validates `aud == <itself>` per RFC 7519 §4.1.3 would *reject* a `aud = issuer` token. Today these tokens validate only because audience checking is optional — `pkg/authjwt` enforces `aud` exclusively when a deployment configures an expected `Audience` (`pkg/authjwt/verifier.go:58-60, 163`). In effect the tokens lean on lax or issuer-configured verifiers.
- It makes the single-audience direction we discussed on the WIMSE thread *nominal* rather than meaningful — there is no real audience to scope down to.

## Proposed change

Adopt [RFC 8707 — Resource Indicators for OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc8707):

1. Accept a `resource` parameter (and optionally `audience`) on `/oauth2/token` and the token-exchange grant.
2. Validate it against an allowed-resource list (per tenant/client).
3. Set `aud` to the requested resource(s) instead of the issuer.
4. Decide the no-`resource` behaviour explicitly — reject, or fall back to a configured default — rather than silently substituting the issuer.

Scope this to the OAuth access-token profile; JWT-SVID / WIMSE assertion paths keep their own audience semantics.

## References

- RFC 8707 — Resource Indicators for OAuth 2.0
- RFC 7519 §4.1.3 — `aud`
- RFC 9068 §3 — `aud` on `at+jwt` access tokens (relates to #189)
- Reported by Andrii Deinega via WIMSE WG review of ZeroID.

saucam:
## Summary

- Adds `resource` parameter support on `POST /oauth2/token` per RFC 8707 §2, stamping the supplied URI as `aud` on the issued access token
- Adds `allowed_resources` and `default_audience` config fields; when `allowed_resources` is non-empty, any `resource` not in the list is rejected with `invalid_target` (fail-closed)
- Propagates `ResolvedAudience` through all six OAuth grant types: `client_credentials`, `jwt_bearer`, NHI token exchange, API key, authorization code, refresh token, and CIBA redeem
- Preserves `audienceScopeProfiles` (BR-7): named audience profile wins over `resource` on external-principal exchange
- Backward-compatible: no `resource` + no `default_audience` → `aud` remains the issuer URL (existing behavior)

## Test plan

- [x] `go test ./internal/service/... -run TestValidateAllowedResources` — 5 sub-tests pass
- [x] `go test ./tests/integration/... -run TestResourceIndicator` — 8 sub-tests pass (open-mode, restricted-mode, blocked, malformed, fragment, no-resource fallback, DefaultAudience, jwt_bearer)
- [x] `go test ./tests/integration/... -run TestIssuedToken` — existing + new `TestIssuedTokenResourceOverridesAud` pass
- [x] Full `go build ./...` clean

Fixes: highflame-ai/zeroid#199


saucam:
Auto-fixed by pr-shepherd (iteration 1 - branch hygiene):
  - Rebased on origin/main: 2 commits (the codeoid audience pipeline-scopes fixes) were already merged upstream and dropped by git during rebase. Branch now has 3 commits ahead of main: the RFC 8707 feature commits.
  - No commit subjects needed rewriting — all 3 match the required CI regex (feat: prefix, no scope parens).

Re-running CI. These are mechanical changes with no behavior impact.

rsharath (approved):

PR #253, opened 2026-07-28
fix: create-api-key 500s (uuid 22P02) when product and identity_id both omitted

### LINKED ISSUE #149: fix: create-api-key 500s (uuid 22P02) when product and identity_id both omitted
## Summary

`POST /api/v1/api-keys` returns **HTTP 500 "failed to create API key"** when the request body omits both `product` and `identity_id`. The underlying error is a Postgres type failure — an empty string is written into the `uuid` column `api_keys.identity_id`:

```
failed to store API key: failed to create API key:
ERROR: invalid input syntax for type uuid: "" (SQLSTATE=22P02)
```

This is the documented minimal request shape (only `name` is `required:"true"` in `CreateAPIKeyInput`), so the simplest valid-looking call fails with an opaque 500.

## Reproduction

```bash
curl -s -X POST http://localhost:8899/api/v1/api-keys \
  -H 'Content-Type: application/json' \
  -H 'X-Account-ID: acme' -H 'X-Project-ID: prod' -H 'X-User-ID: admin' \
  -d '{"name":"my-service-key"}'
```

Response:
```json
{ "title": "Internal Server Error", "status": 500, "detail": "failed to create API key" }
```

Server log:
```
{"level":"error","error":"failed to store API key: failed to create API key: ERROR: invalid input syntax for type uuid: \"\" (SQLSTATE=22P02)","caller":".../internal/handler/apikey.go:147","message":"failed to create API key"}
```

Workaround: include a `product` (e.g. `"product":"my-svc"`) so a service identity is auto-provisioned, or pass an explicit `identity_id`.

## Root cause

`internal/service/apikey.go` → `CreateKey`:

- An identity is auto-provisioned **only** when `req.IdentityID == "" && req.Product != ""` (line ~79). With both empty, `req.IdentityID` stays `""`.
- The key struct is built with `IdentityID: req.IdentityID` (line ~184). `domain.APIKey.IdentityID` is tagged `bun:"identity_id,type:uuid"` (`domain/apikey.go:36`), and the column is `identity_id UUID REFERENCES identities(id) ON DELETE SET NULL` — **nullable** (`migrations/001_init_schema.up.sql:96`).
- `s.repo.Create` then inserts `""` into a `uuid` column → `SQLSTATE 22P02`.
- The handler maps any create error to a generic `huma.Error500InternalServerError` (`internal/handler/apikey.go:147`).

The doc comment on `CreateKey` states *"Every key is linked to an identity and assigned a credential policy"*, but that invariant is enforced only on the `Product != ""` path. The no-product / no-identity path violates it and fails at the DB layer instead of being handled.

## Expected behavior

One of (in rough order of preference):

1. **Persist `NULL`, not `""`** — the column is nullable (`ON DELETE SET NULL`), so an unlinked key should store SQL `NULL` for `identity_id`. Likely needs `IdentityID` to be `*string`/`sql.NullString` (or a `nullzero` bun tag) so empty maps to `NULL`.
2. **Auto-provision in the no-product path too**, if the "every key has an identity" invariant is meant to be hard — derive a default/service identity even when `product` is unset.
3. **Fail with `400`** and a clear message if `identity_id`/`product` is genuinely required, rather than a 500 from a DB type error.

At minimum, this should not surface as an opaque HTTP 500.

## Environment

- zeroid `main` (current `HEAD`), standalone docker-compose, Postgres 17.
- Reproduced against `/api/v1/api-keys` with default tenant headers; no auth/network specifics involved.

safayavatsal:
Closes #149.

## Summary

`POST /api-keys` with only the documented-required `name` field (no `product`, no `identity_id`) returned HTTP 500. `CreateKey` only auto-provisions an identity when `Product != ""`; with both empty, `IdentityID` stayed `""`, and the store inserted that empty string into the `identity_id` `uuid` column — Postgres rejects it (`SQLSTATE 22P02: invalid input syntax for type uuid`). The column is nullable (`REFERENCES identities(id) ON DELETE CASCADE`, no `NOT NULL`), so the schema already supports an unlinked key — the code just wasn't persisting "unlinked" correctly.

## Fix

Per the issue's own preferred option: **persist `NULL`, not `""`**.

- `domain/apikey.go` — `IdentityID` gets the `nullzero` bun tag, matching the identical pattern already used on `ReplacedBy` in the same struct for the same class of optional-UUID-FK field. Empty now persists as SQL `NULL` instead of `""`.

## Defense in depth

While in this code path, also closed the "at minimum, this should not surface as an opaque 500" ask:

- `internal/service/errors.go` — new `isInvalidUUIDError` (SQLSTATE 22P02) and `isForeignKeyViolation` (23503) helpers, following the exact convention of the existing `isDuplicateKeyError` (23505) helper used elsewhere in this package.
- `internal/service/apikey.go` — new `ErrInvalidAPIKeyReference` sentinel. Both `CreateKey`'s `repo.Create` error path and its identity subset-check lookup (`GetIdentity`) now classify DB-level bad-reference errors into this sentinel instead of leaking a raw error.
- `internal/handler/apikey.go` — maps `ErrInvalidAPIKeyReference` to 400, alongside the existing credential-policy error mappings.

**Adjacent bug found in the same pass:** the identity subset-check lookup had the identical failure mode from the other direction — an explicitly-supplied `identity_id` that's malformed or doesn't resolve in the caller's tenant also 500'd, for the same "no `errors.Is` handling" reason. Fixed with the same sentinel, no new abstraction.

## Tests

`tests/integration/apikey_test.go` — three new tests:
- `TestCreateAPIKey_NoProductNoIdentity_Succeeds` — the exact repro from the issue (now 201, empty `identity_id` in the response).
- `TestCreateAPIKey_NonexistentIdentityIDRejected` — a well-formed but nonexistent `identity_id` (400).
- `TestCreateAPIKey_MalformedIdentityIDRejected` — a malformed (non-UUID) `identity_id` (400).

## Verification

Confirmed the `nullzero` fix is load-bearing and not redundant with the error-classification hardening: temporarily reverted just the tag and reran the primary test — it failed. With the error classification alone in place, the failure had already improved from 500 → 400, but the request that should succeed per the documented contract (only `name` is required) still didn't succeed until the tag was restored. Both pieces are independently necessary.

- `GOEXPERIMENT=jsonv2 go build ./...` clean
- `GOEXPERIMENT=jsonv2 go vet ./...` clean
- `gofmt -l .` clean
- Full test suite green: unit tests with `-race`, full `tests/integration` (testcontainers) suite including the 3 new tests and the existing api-key regression tests

### DIFF
--- a/domain/apikey.go
+++ b/domain/apikey.go
@@ -33,7 +33,7 @@ type APIKey struct {
 	KeyVersion         int             `bun:"key_version"              json:"key_version"`
 	AccountID          string          `bun:"account_id"               json:"account_id"`
 	ProjectID          string          `bun:"project_id"               json:"project_id"`
-	IdentityID         string          `bun:"identity_id,type:uuid"    json:"identity_id"`
+	IdentityID         string          `bun:"identity_id,type:uuid,nullzero" json:"identity_id"`
 	CreatedBy          string          `bun:"created_by"               json:"created_by"`
 	Scopes             []string        `bun:"scopes,array"             json:"scopes"`
 	Product            string          `bun:"product"                  json:"product"`
--- a/internal/handler/apikey.go
+++ b/internal/handler/apikey.go
@@ -145,6 +145,11 @@ func (a *API) createAPIKeyOp(ctx context.Context, input *CreateAPIKeyInput) (*Cr
 		if errors.Is(err, service.ErrPolicySubsetViolation) {
 			return nil, huma.Error400BadRequest(err.Error())
 		}
+		// Malformed or dangling reference field (e.g. a nonexistent
+		// identity_id) is caller error, not a server fault — see #149.
+		if errors.Is(err, service.ErrInvalidAPIKeyReference) {
+			return nil, huma.Error400BadRequest("invalid identity_id or credential_policy_id")
+		}
 		log.Error().Err(err).Str("name", input.Body.Name).Msg("failed to create API key")
 		return nil, huma.Error500InternalServerError("failed to create API key")
 	}
--- a/internal/service/apikey.go
+++ b/internal/service/apikey.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"crypto/rand"
 	"crypto/sha256"
+	"database/sql"
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"time"
 
@@ -16,6 +18,12 @@ import (
 	"github.com/highflame-ai/zeroid/domain"
 	"github.com/highflame-ai/zeroid/internal/store/postgres"
 )
+
+// ErrInvalidAPIKeyReference is returned when a caller-supplied reference
+// field (e.g. identity_id, credential_policy_id) fails a database-level type
+// or foreign-key check that request validation didn't catch — malformed
+// input, not a server fault. Callers map this to 400.
+var ErrInvalidAPIKeyReference = errors.New("invalid reference on API key")
 
 // APIKeyService handles CRUD operations for API keys (zid_sk_* keys).
 type APIKeyService struct {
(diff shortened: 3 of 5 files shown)


rsharath:
Superseded by #307, which landed the same fix (400 on missing identity link). Closing as stale.

PR #257, opened 2026-08-05
fix: stamp the RFC 8707 resource binding as its own claim (shield#366)

saucam:
First of three PRs closing highflame-ai/highflame-shield#366. Spec: highflame-ai/highflame-architecture#258.

## Problem

`idJAGBearer` audience-restricts the minted token by writing the ID-JAG's `resource` claim into `aud` (`oauth_id_jag.go:284`) and dropping the name. Downstream that's unrecoverable, because ZeroID stamps `aud` on **every** token it issues — defaulting to the issuer URL to satisfy JWT-SVID §3 (`credential.go:436-440`):

```go
aud := req.Audience
if len(aud) == 0 {
    aud = []string{s.issuer}   // <- indistinguishable from a real restriction
}
```

Shield read a non-empty `aud` as "resource-bound" and hard-denied every MCP-targeted guard request on the JWT path in prod:

```
{"target_server":"slack","token_audience":["https://auth.highflame.ai"],
 "msg":"audience binding denied request before Cedar"}
```

## Change

Carry the binding under its own name so presence of the claim — never the shape of a value — is the discriminator.

| | before | after |
|---|---|---|
| ID-JAG token | `aud: [".../mcp/github"]` | `aud: [".../mcp/github"]`, `resource: [".../mcp/github"]` |
| every other token | `aud: ["https://auth..."]` | unchanged, no `resource` |

- **`oauth_id_jag.go`** — stamp `resource` alongside `aud`. `aud` is untouched, so RFC 8707 §2 conformance and any correct `aud` validator are unaffected. Set *after* the `PropagateClaims` loop so a deployer listing `"resource"` in `propagate_claims` can't clobber the validated value.
- **`oauth.go`** — `resource` joins `reservedClaims`. This one is load-bearing: Shield gates enforcement on presence of the claim as proof the *mint* recorded a binding, and that only holds while ZeroID is the sole party able to set it. Unreserved, a caller could forge a binding on a token that has none, or widen a real one by overriding it through `additional_claims`.
- **`pkg/authjwt`** — typed `Resource []string` so consumers gate on it without reaching into `Custom`. Accepts both RFC 8707 shapes (array, single string); an unexpected shape degrades to *bound* rather than silently to *unbound*, since empty means enforce nothing.

## Tests

`pkg/authjwt/resource_claim_test.go` — array shape, single-string shape, and the shield#366 regression: a token **with** `aud` and **without** `resource` must leave `Resource` empty.

`revocation_test.go` — extends the existing `TestReservedClaims_*` pattern to assert `resource` is reserved.

The "the mint actually stamps it" assertion goes in the regression suite, extending `tests/api/admin/test_id_jag_mas.py`, which already exercises the real mint end-to-end and already asserts `aud == resource`. The repo-local ID-JAG tests are all failure-path and never reach a successful mint.

`go build ./...`, `go vet`, full `internal/service` and `pkg/authjwt` suites green (`GOEXPERIMENT=jsonv2`).

## Next

Needs a `pkg/authjwt` tag before Shield can consume it — Shield is currently pinned to `v1.4.0`, latest is `v1.8.0`.

Refs highflame-ai/highflame-shield#366

### DIFF
--- a/internal/service/oauth.go
+++ b/internal/service/oauth.go
@@ -127,6 +127,13 @@ const defaultExternalPrincipalRefreshTokenTTL = 12 * 3600
 var reservedClaims = map[string]bool{
 	// RFC 7519 registered claims
 	"iss": true, "sub": true, "aud": true, "exp": true, "nbf": true, "iat": true, "jti": true,
+	// RFC 8707 resource binding. Shield reads `resource` as proof that ZeroID
+	// recorded a resource restriction AT THE MINT and enforces INV-IDN-006 on
+	// it — so the claim only means that if ZeroID is the only party who can set
+	// it. Reserving it keeps the presence of the claim load-bearing: a caller
+	// must never be able to make a token look resource-bound when it isn't, or
+	// to widen a real binding by overriding it through additional_claims.
+	"resource": true,
 	// ZeroID identity claims
 	"account_id": true, "project_id": true, "user_id": true, "owner_user_id": true,
 	"external_id": true, "identity_type": true, "sub_type": true, "trust_level": true,
--- a/internal/service/oauth_id_jag.go
+++ b/internal/service/oauth_id_jag.go
@@ -214,6 +214,20 @@ func (s *OAuthService) idJAGBearer(ctx context.Context, req TokenRequest) (*doma
 		}
 	}
 
+	// RFC 8707 resource binding (D4), stamped as its own claim so downstream
+	// PEPs have an EXPLICIT signal that this token was resource-bound at the
+	// mint. `aud` carries the same value for RFC 8707 §2 conformance, but it
+	// cannot serve as the signal: ZeroID stamps `aud` on every token it issues,
+	// defaulting to the issuer URL to satisfy JWT-SVID §3, so a non-empty `aud`
+	// says nothing about whether a binding exists. Shield treating it as if it
+	// did denied every MCP-targeted request in prod (shield#366). Presence of
+	// THIS claim is the discriminator; its absence means "not resource-bound".
+	//
+	// Set after the PropagateClaims loop so a deployer that lists "resource" in
+	// propagate_claims cannot clobber the value we validated above, and
+	// reserved in reservedClaims so additional_claims cannot inject or widen it.
+	customClaims["resource"] = resources
+
 	// Resolve the identity's credential policy when we have a real identity row
 	// so issuance enforces the same policy ceiling (allowed scopes, max TTL,
 	// trust level, policy expiry) as every other grant.
--- a/internal/service/oauth_id_jag.go
+++ b/internal/service/oauth_id_jag.go
@@ -40,6 +40,64 @@ func isIDJAGAssertion(assertion string) bool {
 	return hdr.Typ == idJAGTyp
 }
 
+// buildIDJAGCustomClaims assembles the custom claim set for an ID-JAG-minted
+// access token. Extracted from idJAGBearer so the claim set — notably the
+// RFC 8707 resource binding, which is security-load-bearing — is unit-testable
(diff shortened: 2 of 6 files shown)


abhijitjavelin (approved):

saucam:
## Deep review of this change — findings

Reviewed for regressions across every surface this touches. Three things fixed on-branch (a86e890), one pre-existing hazard worth knowing about, and the rest verified clean.

### Fixed on this branch

**1. The `resource` stamp had no test — and it fails quiet.**

The repo's ID-JAG tests are all fail-closed and never reach issuance; the regression suite can't mint an ID-JAG without the fake-IdP fixture. So the one claim INV-IDN-006 enforcement depends on was landing untested. If it silently stopped being stamped, Shield wouldn't error — it would just stop enforcing.

Extracted `buildIDJAGCustomClaims` and unit-tested it. **Mutation-checked** that the tests actually bite: moving the stamp before the `PropagateClaims` loop makes the clobber test fail with the widened set `["github","slack"]` instead of the validated `["github"]`.

That ordering turns out to be genuinely load-bearing and had rested only on a comment. A deployer listing `"resource"` in `propagate_claims` would otherwise overwrite the `extractResourceClaim`-validated value with the **raw** one — bypassing RFC 8707 shape normalization *and* the fail-closed empty check. That's a real privilege-widening path, now proven closed.

**2. My `Resource` doc comment overclaimed.** It said an unexpected shape "degrades to bound rather than unbound." A probe says otherwise:

| shape | result |
|---|---|
| `["https://gw/mcp/github"]` | bound |
| `"https://gw/mcp/github"` | bound |
| `[]` / `""` | unbound |
| `42` / `{...}` | **unbound — fail open** |
| `["https://gw/mcp/github", 7]` | partially bound (non-strings dropped) |

Fail-open on a malformed shape is tolerable here — ZeroID is the sole issuer behind this verifier, always writes `[]string`, and the value is signature-verified — but it isn't the general property I'd written. Reworded to state the actual limit.

**3.** gofmt alignment in `knownKeys`.

### Pre-existing, not introduced here, but a landmine for zeroid#258

**The refresh grant strips the binding.** `refreshToken` re-mints with no `Audience` and no `CustomClaims` — it inherits only scopes and mission. So refreshing a resource-bound token would yield an **unbound** one.

Not exploitable today: only `authorizationCode` and the external-principal exchange issue refresh tokens, and the ID-JAG path issues none. But the moment zeroid#258 adds a `resource` parameter to the token endpoint, resource-bound tokens start flowing through grants that *do* refresh, and the binding silently drops — a clean escalation. The refresh-token row has no column to carry it, so this needs schema work. Noted on that issue rather than papered over here.

### Verified clean

- **`additional_claims` regression** — reserving `resource` silently drops caller input, so I checked every caller across 9 repos. Only `highflame-admin` uses `additional_claims`, and only for `gateway_id`. No caller sends `resource`.
- **Both `reservedClaims` filter sites** (`oauth.go:954`, `oauth_external_idp.go:166`) share the one map, so the addition covers both.
- **`Custom` map** — `resource` moving to a typed field removes it from `Custom`. Nothing anywhere reads `GetCustom("resource")`, and no token ever carried the claim, so there's nothing to break.
- **Consumer struct literals** — no positional `authjwt.Claims{...}` literals in any consumer, so adding a field is safe.
- **No `DisallowUnknownFields`** anywhere in the token-decode path, so the new claim can't break strict decoding.
- **`proof.go`'s `Audience`** is a proof-verifier audience, not a resource binding — correctly left alone.
- **Blast radius of the new claim is zero**: ID-JAG isn't configured in any environment, so no token in the wild gains a `resource` claim from this.

`go vet` clean, `gofmt` clean, full `./internal/...` and `pkg/authjwt` suites green.

PR #269, opened 2026-08-07
fix(middleware): enforce DPoP sender-constraint on bound agent tokens

SashaMIT:
## Problem

`AgentAuthMiddleware` verified ES256 signature, issuer, and exp — but never read the token's `cnf.jkt` claim and never required a `DPoP` proof header. Meanwhile `credential.go:479` mints `cnf.jkt` whenever a DPoP proof accompanies the token request, and `pkg/dpop` ships `ValidateBoundToToken` (documented as *"use this on resource servers where the access token claim must match the presented proof key"*) with **zero production callers**.

Net effect: DPoP binding existed at issuance and was silently dropped at verification. A leaked bound token replayed as a plain `Bearer` — no proof of key possession — which is precisely the stolen-token scenario DPoP exists to defeat (RFC 9449 §6.1).

## Fix

When the validated token carries `cnf.jkt` and a `DPoPVerifier` is configured:

- Require the `DPoP` header (401 `invalid_token` otherwise).
- Validate the proof via the existing `ValidateBoundToToken`: `htu` from `EffectiveRequestURL` (the middleware already in place for this), `ath` against the presented token, thumbprint match against `cnf.jkt`, replay protection via the shared store.

Unbound tokens (no `cnf.jkt`) are unaffected — this only closes the gap for tokens ZeroID already chose to bind.

## Tests

New `agent_auth_dpop_test.go`: unbound token passes without a proof header; bound token without a proof header is rejected. `go build ./...` and the middleware tests pass (note: repo requires `GOEXPERIMENT=jsonv2`, per the Makefile).

Made with Cursor

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/middleware/agent_auth.go
+++ b/internal/middleware/agent_auth.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/lestrrat-go/jwx/v4/jwa"
 	"github.com/lestrrat-go/jwx/v4/jwt"
+
+	"github.com/highflame-ai/zeroid/pkg/dpop"
 	"github.com/rs/zerolog/log"
 
 	"github.com/highflame-ai/zeroid/internal/jwtalg"
@@ -25,6 +27,9 @@ type AgentClaims struct {
 	Scopes     []string
 	JTI        string
 	IdentityID string
+	// DPoPJKT is the cnf.jkt thumbprint the token was bound to at issuance,
+	// empty for unbound tokens.
+	DPoPJKT string
 }
 
 type agentClaimsKey struct{}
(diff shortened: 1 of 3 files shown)


rsharath:
Cross-reference from the SDK side (highflame-ai/highflame-sdk#105): the SDKs are about to send bound tokens as `Authorization: Bearer <token>` plus a `DPoP: <proof with ath>` header, which is exactly the shape this PR validates, so no conflict here.

One request on ordering, tracked in #322: please keep this PR at "accept Bearer, validate the proof when `cnf.jkt` is present" and do **not** add RFC 9449 §7.2 rejection of bound-as-Bearer in the same change. The SDKs flip to the `DPoP` scheme only after ZeroID accepts it; rejecting Bearer before that ships would 401 every SDK client with `dpop_key` set. #322 has the three-step sequence and the config flag for step 3.

SashaMIT:
Thanks. This PR already keeps Bearer and only validates the DPoP proof when cnf.jkt is present. It does not reject bound-as-Bearer. That stays for #322.

PR #277, opened 2026-08-09
fix: reject empty owner in owner-scoped cascade revocation (closes #275)

### LINKED ISSUE #275: RevokeAllActiveForOwner: empty owner_user_id cascade-revokes every ownerless identity (and the feature is unreachable/untested)
Found while reviewing #270; unrelated to that PR. Both problems are in #273 (`692965a`) as merged to `main`, and nothing has landed after it.

Two coupled problems: the primitive is unsafe with an empty owner, and it is currently unreachable — which is the only reason the first one has not bitten yet.

## 1. An empty `owner_user_id` cascade-revokes every ownerless identity in the account

`internal/service/credential.go:642` defaults the reason and nothing else:

```go
func (s *CredentialService) RevokeAllActiveForOwner(ctx context.Context, ownerUserID, accountID, reason string) (int64, error) {
	if reason == "" {
		reason = "owner_deactivated"
	}
	revoked, err := s.repo.RevokeAllActiveForOwner(ctx, ownerUserID, accountID, reason)
```

The SQL does not filter it either (`migrations/041_revoke_credentials_by_owner.up.sql:50`):

```sql
WHERE i.owner_user_id = p_owner_user_id
  AND i.account_id    = p_account_id
```

So `RevokeAllActiveForOwner(ctx, "", accountID, …)` selects every identity whose `owner_user_id` is `''`, then walks the `parent_jti` chain and revokes each one's whole delegation subtree.

**Empty owner is a real, populated value — not a can't-happen.** `internal/service/identity.go:108` states ownership is *"OPTIONAL — ownerless"* for `discovered` identities, and `internal/store/postgres/identity.go:190` has an explicit ownerless query facet built on exactly that value:

```go
q = q.Where("(owner_user_id IS NULL OR owner_user_id = '')")
```

The realistic trigger is the intended one: an IdP offboarding webhook whose payload is missing or blank in the user-id field. Instead of revoking one departing human's credentials, it revokes every discovered/detached agent in the tenant. That is a tenant-wide outage of exactly the workloads nobody is watching.

The identity-scoped sibling is protected only by accident — its `uuid` cast rejects `''` with a 22P02 rather than by any deliberate guard.

### Suggested fix

Guard in the service:

```go
if ownerUserID == "" || accountID == "" {
	return 0, fmt.Errorf("RevokeAllActiveForOwner requires a non-empty owner and account")
}
```

…and mirror it in the SQL (`AND p_owner_user_id <> ''`) so the guard survives a future caller that bypasses the service. Defense in depth is cheap here and the failure mode is not recoverable — revoked credentials cannot be un-revoked.

## 2. The feature is unreachable and untested

`grep -rn RevokeAllActiveForOwner --include='*.go' .` returns only the definition and the service→repo call. Specifically:

- **No caller.** Nothing invokes it.
- **No export.** No `Server` method and no `hooks.go` wrapper, and it lives under `internal/`, so embedders (highflame-authn) cannot reach it either.
- **No test.** `revoke_credentials_by_owner` appears in `internal/store/postgres/credential.go` only — in a comment and in the query string. Never in a test.

Two consequences:

1. **Human offboarding revokes nothing today.** The stated purpose of #273 is not achieved by anything that runs.
2. **Migration 041 ships with zero coverage.** It is a recursive CTE with `CYCLE` detection, a depth cap, and a data-modifying CTE with `RETURNING` — the most intricate SQL in the repo. The 031 cascades it mirrors are covered; this one is not, so the first execution in anger will also be its first execution.

Wiring a caller without first adding the guard from part 1 is the dangerous ordering. Please do both in the same change.

## Verification

- `origin/main` tip is `692965a` (#273 itself) — nothing has landed after it.
- None of the 12 open PRs touch `internal/service/credential.go`, `internal/store/postgres/credential.go`, or `migrations/041_*`.

saucam:
Closes #275.

## The bug

`RevokeAllActiveForOwner` accepted an empty `owner_user_id` and passed it straight into migration 041's seed predicate, which is a plain equality:

```sql
WHERE i.owner_user_id = p_owner_user_id
  AND i.account_id    = p_account_id
```

`identities.owner_user_id` is `VARCHAR(255) NOT NULL` (001_init_schema), so an ownerless identity stores `''` — **not** NULL. And ownerless is a documented posture for `discovered` identities, not an anomaly: `identity.go` calls ownership *"OPTIONAL — ownerless"* for them, and `postgres/identity.go` has an explicit `(owner_user_id IS NULL OR owner_user_id = '')` facet built on exactly that value.

So a blank owner matched **every ownerless identity in the account** and cascade-revoked each one's entire delegation subtree.

The realistic trigger is the intended caller: an IdP offboarding webhook whose user-id field arrives blank turns *"revoke one departing human's agents"* into a tenant-wide outage of precisely the workloads nobody is watching. Revoked credentials cannot be un-revoked.

## Reproduced against a real database

With migration 042 removed, seeding two ownerless identities plus a delegated child and calling with a blank owner:

```
WITHOUT MIGRATION 042 -> err=<nil>, revoked 3 credential(s):
  [ownerless-a-...-jti  ownerless-b-...-jti  ownerless-a-child-...]
```

No error, three credentials gone — including the delegated child, so the cascade reached past the seed rows. With 042 the function raises and nothing is touched.

## The fix — guards at both layers

They protect different callers, and the failure mode is unrecoverable, so both earn their place:

- **`internal/service/credential.go`** rejects an empty `owner_user_id` or `account_id` before reaching the repo.
- **Migration 042** mirrors it with `RAISE EXCEPTION` so the guarantee survives a caller that bypasses the service — a psql session, another service, or a repo method added later.

`RAISE` rather than returning zero rows on purpose: a blank owner is a caller bug, and silently reporting *"revoked 0 credentials"* would let a broken offboarding integration look healthy indefinitely.

**041 is untouched** — it's already merged, so 042 `CREATE OR REPLACE`s the function, mirroring how 031's down restores the 029 bodies. The down migration restores 041's body and carries an explicit warning that rolling back reintroduces the hole.

Everything else in the function — the recursive walk, CYCLE guard, depth cap 50, liveness filters on the final UPDATE only, and the RETURNING projection — is carried over verbatim. Signature unchanged, so no caller changes.

## Tests — migration 041's first coverage

041 shipped with none, which is the second half of #275. It's the most intricate SQL in the repo (recursive CTE + CYCLE detection + depth cap + data-modifying CTE with RETURNING), and its first execution in anger would have been its first execution ever.

`tests/integration/revoke_by_owner_test.go` (real Postgres):
- empty owner rejected; both ownerless credentials **and** the delegated child survive — **verified to fail without 042**
- empty account rejected (the tenant-scope half)
- happy path: owner's credential + child + grandchild revoked exactly once each, while another owner's credential and an ownerless identity in the same account are untouched

`internal/service/credential_revoke_owner_test.go`: the service guard, built with a **nil repo deliberately** — a panic there proves the guard failed to short-circuit before the DB layer. Mutation-checked: removing the guard makes them fail.

`go vet` clean, `gofmt` clean, unit suite green, full integration suite green (68s).

## Not in this PR

#275 also notes the primitive is **unreachable** — no caller, no `Server` method, no `hooks.go` wrapper. I've deliberately left that out: the identity-scoped sibling isn't exported either (it's reached from internal deactivation flows), so exposing this one means inventing a new public API shape, and choosing the offboarding entry point — Server method? admin endpoint? webhook handler? — is a design decision rather than a bug fix.

The issue's warning was that *"wiring a caller without first adding the guard is the dangerous ordering"*. This PR establishes the guard, so whichever entry point is chosen can now be wired safely. Happy to follow up once we've picked the shape.

### DIFF
--- a/internal/service/credential.go
+++ b/internal/service/credential.go
@@ -639,7 +639,30 @@ func (s *CredentialService) RevokeAllActiveForIdentity(ctx context.Context, iden
 // Fires one RevocationNotifier event per affected JTI after the revocation
 // commits, exactly like RevokeAllActiveForIdentity, so Shield's deny-set picks
 // them up within seconds.
+//
+// Both ownerUserID and accountID are REQUIRED and rejected when empty. This is
+// a hard guard, not defensive tidiness: identities.owner_user_id is NOT NULL
+// (001_init_schema), so an ownerless identity — the documented posture for
+// `discovered` ones (identity-lifecycle.md "Ownership: relaxed for discovered
+// only") — stores the empty string. An empty ownerUserID would therefore match
+// EVERY ownerless identity in the account and cascade-revoke each one's whole
+// delegation subtree. The realistic trigger is the intended caller: an IdP
+// offboarding webhook whose user-id field arrives blank, turning "revoke one
+// departing human's agents" into a tenant-wide outage of exactly the workloads
+// nobody watches. Revocation is not reversible, so this fails closed.
+//
+// The identity-scoped sibling is protected only incidentally — its uuid cast
+// rejects "" with a 22P02 — so the guard lives here explicitly rather than
+// relying on that accident holding for a TEXT column. Mirrored in SQL by
+// migration 042 so it survives a caller that bypasses this service.
 func (s *CredentialService) RevokeAllActiveForOwner(ctx context.Context, ownerUserID, accountID, reason string) (int64, error) {
+	if ownerUserID == "" || accountID == "" {
+		return 0, fmt.Errorf(
+			"RevokeAllActiveForOwner requires a non-empty owner_user_id and account_id (got owner=%q account=%q): "+
+				"an empty owner matches every ownerless identity in the account",
+			ownerUserID, accountID)
+	}
+
 	if reason == "" {
 		reason = "owner_deactivated"
 	}
--- a/internal/service/credential_revoke_owner_test.go
+++ b/internal/service/credential_revoke_owner_test.go
@@ -1,0 +1,66 @@
+package service
+
+import (
+	"context"
+	"testing"
+)
+
+// TestRevokeAllActiveForOwner_RejectsEmptyOwner pins the guard that keeps a
+// blank owner from cascade-revoking every ownerless identity in an account.
+//
+// Why this matters: identities.owner_user_id is VARCHAR(255) NOT NULL, so an
+// ownerless identity stores the empty string rather than NULL — and ownerless is a documented
+// posture for `discovered` identities, not an anomaly. Migration 041's seed
+// predicate is a plain equality on that column, so an empty owner selects every
+// one of them and walks each delegation subtree. The realistic trigger is an IdP
+// offboarding webhook whose user-id field arrives blank.
+//
+// The receiver is built with a nil repo ON PURPOSE. The guard must short-circuit
+// before any repository call, so a nil-pointer panic here is a real failure
+// signal: it means the guard did not fire and execution reached the DB layer.
+func TestRevokeAllActiveForOwner_RejectsEmptyOwner(t *testing.T) {
+	t.Parallel()
+
+	svc := &CredentialService{} // repo intentionally nil — see doc comment
+	ctx := context.Background()
+
+	tests := []struct {
+		name      string
+		ownerUser string
+		accountID string
+	}{
+		{"empty owner", "", "acct-123"},
+		{"empty account", "user-42", ""},
+		{"both empty", "", ""},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			n, err := svc.RevokeAllActiveForOwner(ctx, tc.ownerUser, tc.accountID, "owner_deactivated")
+
+			if err == nil {
+				t.Fatalf("expected an error for owner=%q account=%q; a blank owner must never "+
+					"reach the cascade — it matches every ownerless identity in the account",
+					tc.ownerUser, tc.accountID)
+			}
+			if n != 0 {
+				t.Errorf("revoked count = %d, want 0 — nothing may be revoked on the reject path", n)
+			}
+		})
+	}
+}
+
+// TestRevokeAllActiveForOwner_GuardPrecedesReasonDefault documents ordering: the
+// argument guard runs before the reason default, so a caller passing an empty
+// reason AND an empty owner still gets the argument error rather than having the
+// blank owner silently carried into the query with a defaulted reason.
+func TestRevokeAllActiveForOwner_GuardPrecedesReasonDefault(t *testing.T) {
+	t.Parallel()
+
+	svc := &CredentialService{} // nil repo: reaching it would panic
+	if _, err := svc.RevokeAllActiveForOwner(context.Background(), "", "", ""); err == nil {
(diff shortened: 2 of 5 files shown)


rsharath (approved):

saucam:
## Self-audit: does this change introduce the same class of bug, and does the pattern recur?

### This change

- **042's body is byte-identical to 041** apart from the two guard blocks — diffed with comments stripped. The recursive walk, CYCLE guard, depth cap 50, liveness filters and RETURNING projection are carried over verbatim, so `CREATE OR REPLACE` can't silently change traversal semantics.
- **042.down is a faithful revert** — diffs identical to 041's body.
- **The guard can't reject a legitimate call.** There is no valid empty-owner or empty-account invocation: the function's entire purpose is owner-scoped, account-scoped revocation. This is the failure direction that bit #272 (a guard so strict it rejected every legitimate request), so it was worth checking explicitly rather than assuming.
- **No test pollution.** The new integration tests seed identities and live credentials into the shared `testAccountID`. Every other account-scoped counter in the suite is keyed by a specific `id` / `identity_id` / `credential_policy_id`, none count account-wide, and the full suite passes green (68s).
- **Both suites mutation-checked** — removing the service guard panics on the nil repo; removing migration 042 makes the empty-owner test fail with 3 credentials revoked.

### Does the pattern recur elsewhere?

Swept every SQL function and every bulk destructive repo operation. The shape that matters is *"a legitimately-empty string is the primary breadth selector"*.

| Site | Selector | Verdict |
|---|---|---|
| `revoke_credentials_cascade` | `p_identity_id UUID` | Safe — uuid cast rejects `''` (22P02) |
| `revoke_credential_cascade` | `p_id UUID` + TEXT account/project | Safe — the UUID PK narrows to one row; empty TEXT can only narrow further, never widen |
| `SigningCredential.RevokeWorkload` | `workload` TEXT, **no service-layer validation** | Safe **by upstream construction** — see below |
| `IdentityRepository.DeactivateStaleDiscovered` | `source_id`, documented as legitimately empty | Same shape, materially lower severity — see below |

**`RevokeWorkload`** has the identical shape to the pre-fix code here: `WHERE workload = ?` with a `VARCHAR(255) NOT NULL` column and a service method that is a pure passthrough with no validation. It's safe only because `Attest` — the sole path reaching `SigningCredentialRepository.Create` — rejects an empty/whitespace workload at issuance, so no row can hold `workload = ''` and an empty revoke matches zero rows.

That is fail-closed, but *by accident of upstream validation* rather than by a local guard — precisely the property #275 criticised about the identity-scoped sibling ("protected only by accident"). Not a bug today; worth a symmetric guard if anyone touches that path. I've deliberately not bundled it here, since it isn't currently exploitable and this PR is a targeted fix.

**`DeactivateStaleDiscovered`** does take a legitimately-empty `source_id` (the model documents it as empty for rows ingested without a source). But the blast radius is bounded by `status = 'discovered'` — rows the code notes hold no credentials, so nothing cascades — plus a staleness window, and it's a reversible status flip rather than irreversible revocation. An empty `sourceID` sweeping the "no source" bucket is also arguably the intended semantics. Flagging as an observation, not a defect.

### Conclusion

`revoke_credentials_by_owner` was uniquely vulnerable because it is the **only** function whose breadth is set by a TEXT column that has a legitimately-populated empty value. That property doesn't hold anywhere else in the schema.

Issue #264, opened 2026-08-07
Confidential CIMD clients: parse jwks_uri/jwks and accept private_key_jwt


rsharath:
## Context

Prep for the OpenID AIIM CG **MCP Security Interoperability Event** (Gartner IAM Summit, Dec 7–9 2026). Commitment date **Aug 10 2026**.

**This is the one that decides whether we can check the CIMD box at all.**

## Why it is load-bearing

The event's qualification rule (plan p.11): the CIMD row on the top-level results matrix is checked only if, in the CIMD matrix, an implementation has

- ≥1 row checked in the **Client ID Metadata** section, **and**
- ≥1 row checked in either **Authorization code grant with PKCE** or **Client credentials grant**.

We satisfy the first (`redirect_uris` — enforced, tested, passing). We score **zero** on the second: every row there names a client-authentication method (Client secret / JWT / MTLS), and ZeroID's CIMD clients are public by construction.

So today ZeroID cannot check the CIMD box, despite a working CIMD implementation.

## What is missing

`cimdMetadataDocument` (`internal/service/cimd.go:130`) has no `jwks_uri` and no `jwks` field — the document's key set is never parsed. And `synthesizeCIMDClient` (`internal/service/cimd.go:446`) rejects any document whose `token_endpoint_auth_method` is not `none`:

```go
// Public client only in v1. ... any explicit confidential method is
// rejected — ZeroID does not accept private_key_jwt token-endpoint auth yet.
if authMethod != "none" {
    return nil, fmt.Errorf("%w: token_endpoint_auth_method must be \"none\" (public PKCE); %q is not supported", ...)
}
```

The module docstring already names this as the intended follow-up (`internal/service/cimd.go:25-27`): *"Confidential CIMD clients (private_key_jwt with a published jwks_uri) are future work."*

## Scope

1. Parse `jwks_uri` and `jwks` from the CIMD document.
2. Accept `token_endpoint_auth_method: private_key_jwt` on a CIMD client, verifying the RFC 7523 §2.2 client assertion against the document's key set.
3. Key fetch reuses the existing SSRF-guarded, size-capped, timeout-bounded client the document fetch already goes through, and honours the same cache TTL rules.

**Depends on #206** — that lands `private_key_jwt` client authentication for registered clients (`client_assertion` validation, which does not exist anywhere today). This issue is the CIMD-specific half: sourcing the key set from a self-published document instead of the client registry.

[...]

Issue #263, opened 2026-08-07
RFC 6749 §4.1.2.1: /oauth2/authorize returns JSON errors instead of redirecting them to the client


rsharath:
> ### Scope narrowed
>
> This issue originally covered three things. Two are settled, so it now tracks
> only the third — **RFC 6749 §4.1.2.1 error redirects**. The original text is
> preserved below for context; items 1 and 2 in "What is needed" are done or
> moved.
>
> | Original item | Status |
> |---|---|
> | 1. `GET /oauth2/authorize` with the standard query parameters | **Done** in #270, along with the missing `authorization_endpoint` in AS metadata and the empty-resolver 503, which #270 handles by omitting `authorization_code` from metadata rather than by shipping a resolver |
> | 2. A user-authentication leg (login/consent surface or IdP delegation) | **Moved.** Both halves already exist in-house and the handoff needs no ZeroID change — see highflame-authn#157 and highflame-studio#1392. The mis-scoped ZeroID version was #271, now closed |
> | 3. RFC 6749 §4.1.2.1 error redirects | **Open — this issue** |
>
> ### What remains
>
> Every error path at `/oauth2/authorize` returns a JSON body. All 13 call sites
> go through `writeAuthorizeError`, whose own doc says *"Used for all error paths
> at /oauth2/authorize in v1 (no redirect-with-error)"*, and there is no
> `error=` redirect anywhere in the handler.
>
> RFC 6749 §4.1.2.1 says that once `redirect_uri` has been validated against the
> client's registered list, errors SHOULD go back to it as
> `?error=…&error_description=…&state=…` rather than to the user agent as a
> response body.
>
> That mattered less when the endpoint was POST-only and CLI clients parsed the
> JSON. It matters now that #270 mounts `GET`: on an error the **browser** is
> left looking at a raw JSON body while the client sits waiting on a callback
> that never fires. A partner debugging an interop session against us sees
> "nothing happened" instead of `error=invalid_scope` arriving at their redirect
> URI, which turns a five-second fix into a support thread.
>
> The ordering constraint is the reason this was deferred rather than bundled:
> errors raised **before** `redirect_uri` is validated must stay as JSON — a
> redirect to an unvalidated URI is an open redirect. Only post-validation
> failures (bad scope, rejected client, issuance failure) can be redirected, and
> `redirect_uri` validation currently happens inside `IssueAuthCode`, so the
> handler has no validated URI to redirect to until that call returns. Splitting
[...]

PR #278, opened 2026-08-10
fix: authorization_code off switch, access-log redaction, and CIMD client_id URL conformance

rsharath:
Follow-ups to #270 and #241, both already on `main`. Four fixes: a documented off switch that did not switch anything off, a credential that could reach our own access log, and two CIMD `client_id` rules the draft makes MUST NOT.

> **This PR makes the shipped spec true.** #176 merged the wire-contract spec, whose §12.1 states that userinfo and dot-segment `client_id` URLs are `rejected` — but `main`'s `ResolveClient` enforces neither (the only userinfo check on `main` is on `redirect_uri`, from #241). So `main` currently documents behaviour it does not have, and finding 3 below is what closes that gap.

## 1. `SetAuthorizationCodeAvailable` was an advertisement switch wearing the name of an off switch

It suppressed the discovery metadata and left `/oauth2/authorize` serving.

That gap lands hardest on the deployment the hatch exists for. `AuthorizeRequest.Cookie` predates #270, so a cookie-based `PrincipalResolver` could already be in the field — and it was **safe** while POST was the only route, because `SameSite=Lax` withholds the cookie on a cross-site POST. Mounting GET makes it reachable by cross-site **top-level navigation**, where Lax *does* send the cookie, and CIMD accepts an attacker-published `client_id` carrying its own `redirect_uri`. A deployer who reached for the documented hatch got metadata edits and a live endpoint.

`authorizeHandler` now refuses with **503 on both GET and POST** when the predicate reports false.

**Nothing changes by default.** The built-in guess is "any resolver registered", and a deployment with none already answered 503 here via `ErrNoResolversRegistered` — this reaches the same answer before touching the request. Godoc on both setters and `docs/cimd.md` now say *off*, not *unlisted*.

## 2. A credential could land in ZeroID's own access log

The comment justifying the GET leg claimed a credential *"never lands in a URL, an access log, or a Referer header"*. That was false for our own log: four request loggers recorded `r.RequestURI`, which carries the query string.

A client author who mirrors the documented POST form into a GET gets a correct 401 — *after* the `api_key` has been written to disk and shipped to the deployer's log sink, twice per request (`request.start` and `request.complete`).

The loggers now record `logSafePath(r)`, which is `r.URL.Path`. A path cannot hide a credential, and no ZeroID endpoint takes a query parameter worth logging; a deployer who wants query detail can add their own middleware and choose their own redaction.

The claim is narrowed to what ZeroID can actually keep: **it never *authenticates* from a URL.** Referer and browser history remain the caller's to manage.

## 3. Two CIMD `client_id` URL rules the draft makes MUST NOT

Nothing in the code pins a CIMD draft revision. Checked against the current one — **-02**, July 2026, and #241 merged Aug 1, so it is the right target. These are also the two rows the spec on `main` (§12.1, via #176) already advertises as enforced. The 5 KB document bound matches §8.7, the cache policy sits inside §5.2, and our `§4` self-reference citation is correct. Two §3 rules were not enforced:

| `client_id` | draft-02 §3 | before |
|---|---|---|
| `https://legit.example.com@evil.example/client.json` | userinfo **MUST NOT** | accepted |
| `https://app.example.com/a/../client.json` | dot segments **MUST NOT** | accepted |

**Userinfo is the phishing shape.** It resolves to `evil.example` — `domainAllowed` sees the real host, so this was never an allow-list bypass — while *reading* as `legit.example.com` on a consent screen and in the audit log, the two places a human is asked to trust the string.

**Dot segments** gave one document many spellings, splitting the resolution cache and handing one client several identities the §4 self-reference check cannot distinguish. Checked against the decoded path, so `%2e%2e` is caught too.

Both were confirmed by running the real gate before the fix was written.

We also reject a **query string** where §3 says only SHOULD NOT. That behaviour stays — the `client_id` must remain byte-identical to the URL the document was fetched from, which is exactly what the self-reference check compares — but the comment and docs claimed "the draft requires" and now say what it actually says.

## 4. The CIMD hardening lever ships off, silently

`cimd.allowed_domains` is documented as the primary production hardening lever and ships **empty**, so any public HTTPS host may publish a document that mints a client here. Fair for an open ecosystem, poor for a closed deployment, and invisible either way. ZeroID now warns at startup when CIMD is enabled without an allowlist.

The warning reports the **effective** allow-list, via a new nil-safe `CIMDService.AllowedDomainCount()`. `NewCIMDService` lower-cases and drops blank entries, so the raw config slice and the policy in force disagree — and in the worst direction: `allowed_domains: [""]` has length 1 and an effective count of 0, meaning a YAML typo would leave the deployment wide open while suppressing the warning built to expose that. The `log.Info()` line had the same bug. Both now read one definition of "how many domains are in force".

`docs/cimd.md` also now states what the allowlist *cannot* do. It scopes which hosts may publish, at domain granularity. It does not establish that the party presenting a `client_id` controls that document: CIMD has no proof of possession, so any client may present any published URL — and for a native client whose document lists a loopback `redirect_uri`, the code is delivered to the presenter's own listener, which PKCE does not prevent because the attacker chooses their own verifier. A property of the model rather than a bug here, but it should not be mistaken for client authentication.

## Smaller corrections

* `state` is marked **optional** in the CIMD authorize example. PKCE is mandatory at this endpoint (`code_challenge` and `code_challenge_method` are both in the required-field gate; `state` is not) and carries the CSRF binding `state` originally existed for, so OAuth 2.1 leaves `state` to application state only. Kept in the example, since ZeroID does round-trip it.
* **RFC 6749 §3.1, not §4.1.1**, for the authorization endpoint's HTTP methods. Three sites cited §4.1.1 for "MUST support GET" — that sentence is in §3.1, which is also the one that makes POST a `MAY`: *"The authorization server MUST support the use of the HTTP GET method for the authorization endpoint and MAY support the use of the POST method as well."* OAuth 2.1 §3.1 carries it forward verbatim. The comment now records that mounting both is conformant and that POST stays because it is the only method v1 had. Our remaining §4.1.1 citations are for `response_type` and `state`, which genuinely live there, and are unchanged.
* The header-stub doc comment in `helpers_test.go` had been appended to the *form* stub's comment and sat above the form stub — claiming the function below it read headers and was registered first, when it reads `req.Form` and is registered second. Moved to the resolver it describes.

## Tests

**Endpoint gate (3):** refused on both methods; an available deployment still reaches the handler, so a blanket 503 cannot pass; a nil predicate still means available.

**Access log (1):** the logged path carries neither the credential nor a query string, and is still the bare endpoint an operator reads the line for.

**CIMD `client_id` (5 new cases):** userinfo in both spoof and plain forms, `..`, `.`, and percent-encoded `%2e%2e`.

**CIMD allow-list (2):** a table over nil / empty / `[""]` / whitespace-only / blanks-plus-a-real-domain / case-insensitive duplicates / two distinct domains, plus a behavioural test that a blank-only allow-list really does admit a host no entry names — so the reported count and the enforcement cannot drift apart.

All mutation-checked:

| mutation | result |
|---|---|
| delete the step-0 gate | `= 400, want 503` |
| `logSafePath` → `r.RequestURI` | fails printing the leaked `zid_sk_` value |
| drop the userinfo check | resolver proceeds to an **outbound fetch** of the attacker URL |
| drop the dot-segment check | same |
| constructor keeps blank allow-list entries | `AllowedDomainCount() = 1, want 0` on the `[""]` case |

## Verification

- `go build ./...` and `go vet ./...` clean
- `go test ./...` green across all **9** packages, integration suite included
- `golangci-lint` under CI's exact invocation (`-c .golangci.yml --new-from-merge-base=origin/main`): **0 issues**

## Not addressed here

Deliberately out of scope, with reasons:

- **Invert the gate's default to opt-in.** Inverting silently un-advertises the grant for any deployment that *does* have a GET-capable resolver, trading a loud 401 for a quiet capability loss. The default already suppresses the no-resolver case and is documented as necessary-not-sufficient.
- **GET error paths return JSON instead of redirecting** (RFC 6749 §4.1.2.1). Now filed as **#279**, with the ordering analysis — the blocker is that principal resolution runs *before* any client lookup, so `access_denied` fires with no validated `redirect_uri` to redirect to. The code comment previously pointed at #263, which was narrowed and closed, leaving this untracked; it now points at #279.
- **CIMD `client_id` reuse, and per-installation keys.** Both are properties of the CIMD model, not defects here: there is no proof of possession, and one document per application means one `jwks_uri` for the whole install base, so `private_key_jwt` degrades to a shared credential. The second materially changes what "confidential CIMD client" can mean for a native app — tracked on #264.


internal/handler/authorize.go
```diff
@@ -44,8 +44,16 @@ import (
...
+// header or cookie. ZeroID therefore never AUTHENTICATES from a URL.
+//
+// That is a narrower claim than "a credential never lands in a URL", which
+// an earlier version of this comment made and could not keep: ZeroID does
+// not control what a caller puts in the query string. A client author who
+// mirrors the documented POST form into a GET gets a correct 401 — and
+// would, before this was fixed, have had the secret written to ZeroID's own
+// access log on the way there. The request loggers now record r.URL.Path
+// rather than r.RequestURI for that reason; see logSafePath in server.go.
+// Referer and browser history remain the caller's to manage.
```
adeinega, line 63: Both OAuth 2 and 2.1 don't support a POST request to the authorization endpoint, a client constructs and passes everything as query parameters... unless it's a PAR request.

If your goal is to hide the authorization code it returns from access log files and the history in the browser, you need to pass it as a URI fragment (after `#`), but its receiver needs to be the browser with JS enabled.

OpenID Connect supports both methods. Although, in my practice, POST requests caused more harm and less good.

rsharath: POST is permitted — but our comment cited the wrong section for it, which is likely what prompted the question. Fixed either way.

The governing sentence is **RFC 6749 §3.1**, not §4.1.1:

> The authorization server MUST support the use of the HTTP `GET` method for the authorization endpoint and MAY support the use of the `POST` method as well.

**OAuth 2.1 §3.1** carries that forward verbatim (`draft-ietf-oauth-v2-1-13`, §3.1 ¶4), and as you say OIDC supports both. So GET is an obligation, POST a permitted extra — and PAR is a different mechanism (a POST to a *separate* endpoint that returns a `request_uri` for the GET), not what makes POST here legal.

Three sites in this repo claimed §4.1.1 for the GET requirement. §4.1.1 is the Authorization Request section — it defines the parameters, makes `response_type` REQUIRED and describes `state`, which is what our other §4.1.1 citations are for. Corrected to §3.1, and the comment now records that POST is `MAY`-permitted rather than leaving a reader to wonder.

**Why POST stays:** it is the only method v1 had, so CLI callers depend on it. Removing it is a breaking change with no security benefit. Your instinct that POST causes more harm than good is well taken on one specific axis, though — supporting both methods opens a parameter-smuggling seam, where a caller puts one `client_id` in the body and another in the query and the server reads whichever it happens to reach. #270 closed that by binding each method to exactly one source (`r.PostForm` on POST, `r.URL.Query()` on GET) rather than reading the merged `r.Form`, with `TestAuthorizePOST_QueryParamsAreNotRead` and `TestAuthorizeGET_CredentialInQueryStringIsIgnored` holding both halves.

**On the fragment:** agreed that a fragment is the only way to keep a value out of the URL the receiver logs, and that it needs a JS receiver. It is a different exposure from the one this PR fixes, though. Finding 2 here is about the *request* — a credential the caller put in the query string reaching **ZeroID's own access log**, because our request loggers recorded `r.RequestURI`. The authorization *code* travels in the redirect back to the client's own `redirect_uri`, so it lands in the client's history and logs, not ours; we log path only, never the `Location` header. For that leg the code is single-use and PKCE-bound, so a `response_mode=fragment` option would be a real feature request rather than a fix — happy to file one if you think it is worth having.

rsharath: Fully agreed, and the distinction you draw — `POST /authorize` is still a front-channel request through the user agent, PAR is a direct authenticated client→AS call that moves the payload off the front channel — is the part worth writing down. They are not alternatives to each other.

I went and checked what Highflame actually does as a client, since that is where this advice bites. Both downstream paths in AuthN already build front-channel `GET` with query parameters and neither ever POSTs to an authorization endpoint:

| | `handleConnectStart` | `handleDiscoverStart` |
|---|---|---|
| Endpoint source | hand-configured `authorize_url` | RFC 8414 discovery |
| PKCE | none | S256 |
| Method | `GET` + query | `GET` + query |
| PAR | n/a | **not attempted** |

So the baseline is right by construction. The gap is the second half of your advice: `oauthServerMetadata` parses exactly three fields — `authorization_endpoint`, `token_endpoint`, `registration_endpoint` — so `pushed_authorization_request_endpoint` is never read, and we cannot tell whether an AS offers PAR or requires it.

That last case is a real failure today, and it is the reverse of what you were warning about. RFC 9126 §5 defines `require_pushed_authorization_requests`; against an AS that advertises it `true` we discover metadata, register a client, generate PKCE, hand back an `authorize_url` — and the AS rejects it in the user's browser. Every step on our side succeeded, so nothing in our logs explains it. Filed as highflame-ai/highflame-authn#161, including failing loudly at discovery time rather than in the address bar, and preferring PAR whenever it is offered rather than only when required.

On ZeroID as an AS, for completeness: GET is mandatory and supported, POST is the `MAY` extra, and we implement no PAR — so we correctly advertise no `pushed_authorization_request_endpoint` and no client will try. Whether ZeroID should *offer* PAR is a separate question; happy to file it if you think it belongs on the roadmap, particularly with the interop event in mind.

One adjacent thing I noticed while looking: `handleConnectStart` sends no PKCE at all — `response_type`, `client_id`, `redirect_uri`, `state`, `scope`. That path is platform OAuth apps (GitHub, Slack) where the code exchange is client-secret authenticated, so it is defence-in-depth rather than a hole, and PKCE support there is provider-dependent. Noted as out of scope on #161; say if you would rather it were tracked separately.

server.go
```diff
@@ -304,6 +304,16 @@ func NewServer(cfg Config, opts ...ServerOption) (*Server, error) {
...
+		// cimd.allowed_domains is documented as the primary production
+		// hardening lever, and it ships off. Empty means any public HTTPS host
+		// can publish a document that mints a client here — anyone who can
+		// serve a path on any reachable domain, which inside an org is a low
+		// bar. That is a legitimate posture for an open ecosystem and a poor
+		// one for a closed deployment, and the difference is invisible unless
+		// somebody says so at boot.
+		if len(cfg.CIMD.AllowedDomains) == 0 {
+			log.Warn().Msg("CIMD: cimd.allowed_domains is empty — any public HTTPS host may publish a client_id metadata document. Set an allowlist to run CIMD as a closed ecosystem.")
+		}
```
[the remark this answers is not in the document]
rsharath: Good catch, and correct — verified at `internal/service/cimd.go:215`, which drops any entry that trims to empty. So `allowed_domains: [""]` gives a raw slice of length 1 and an effective allow-list of 0: `domainAllowed` admits every host while the warning stays silent. A YAML typo would suppress precisely the warning that exists to expose an open deployment.

The `log.Info()` line above it had the same bug and would report `allowed_domains=1` for a deployment with no allow-list in force.

Fixed in the latest commit, but at the source rather than by adding a non-blank check in `server.go` — duplicating the trim rule there is how the two drifted apart in the first place. `CIMDService` now exposes `AllowedDomainCount()`, the effective count, nil-safe like `Enabled()`, and both log lines read it. There is now one definition of "how many domains are in force".

Two tests, because the count and the enforcement can drift independently:

- a table over nil, empty, `[""]`, whitespace-only, blanks mixed with a real domain, case-insensitive duplicates, and two distinct domains;
- a behavioural test that a blank-only allow-list really does admit a host no entry names — so a future change cannot fix the count while leaving `domainAllowed` open, or vice versa.

Mutation-checked: making the constructor keep blank entries fails with `AllowedDomainCount() = 1, want 0` on exactly the `[""]` case.

jalbrethsen-highflame (approved):

PR #292, opened 2026-08-12
feat: release a deleted connector's discovered identities so another can adopt

saucam:
## The gap

`reconcileDiscovered` adopts a `source_id` only onto a row that has none:

```go
// Never overwrite a non-empty source_id, so one connector can't claim another's row.
if identity.SourceID == "" && req.SourceID != "" { identity.SourceID = req.SourceID }
```

That guard is correct — but it has **no counterpart for a connector that goes away.**

When a discovery connector is deleted, its discovered rows keep pointing at it forever. The connector never syncs again, and `DeactivateStaleDiscovered` is scoped `WHERE source_id = ?`, so **no prune can reach them**. A replacement connector rediscovering the same agents reconciles onto the same rows (dedup is on `external_id`, so no duplicates) but **cannot take ownership**.

Result: the rows sit as `discovered` indefinitely — presenting as live inventory long after the agent is gone from the source, and un-prunable by anything.

## The change

`POST /identities/discovered/release-source` clears `source_id` on the still-`discovered` rows of one `(origin, source_id)`, so the **existing** adopt-when-empty branch hands them to whichever connector rediscovers them next. No change to reconcile itself.

### Deliberate choices

- **A release, not a deactivation.** Those rows are a human's pending adopt/dismiss queue, and replacing a connector isn't a decision about the agents it found. Rows whose agent is genuinely gone stay `discovered` until dismissed by hand — same as any other discovered row nobody acts on.
- **Adopted and dismissed rows are untouched.** They're outside prune scope, so their `source_id` is *provenance*, not a claim; clearing it would destroy the record of where the identity came from.
- **Scoped like prune.** `origin` and `source_id` are both required, so a release can never unclaim a whole tenant's inventory.
- **The live-connector guard is preserved.** Two live connectors still can't steal each other's rows — only an explicit delete-time release unclaims anything.

## Known residual

Release helps rows the replacement **rediscovers**. An agent that vanished from the IdP *before* the connector was replaced ends up with an empty `source_id` and still isn't reachable by a source-scoped prune — it stays `discovered` until dismissed by hand.

That's not a regression (those rows are un-prunable today too, just bound to a ghost connector instead of unbound), and it's strictly better: an unowned row is honestly unowned, and re-adoptable. Fully closing it would mean either sweeping `source_id = ''` rows — which would also catch manually-ingested identities — or deactivating at delete time, which overrides the human's queue. Neither seemed worth it; happy to be argued out of that.

## Tests

Four integration tests against real Postgres:

- **the whole scenario** — original connector claims a row; a replacement reconciles onto it but *cannot* take ownership (asserting the guard that causes the problem); release; the replacement then adopts it, same row, no duplicate
- **scoping** — a release never unclaims another connector's rows
- **provenance** — an adopted identity is not released and keeps its `source_id`
- **validation** — native origin rejected, missing `source_id` rejected

```
gofmt -l .                                # clean
go vet ./...                              # clean
golangci-lint run ./...                   # 0 issues
go test ./... -race                       # green
go test -tags=integration ./tests/...     # green (43s)
```

## Context

Surfaced by an adversarial review of **highflame-ai/highflame-discovery#50**, which makes delete-then-recreate-under-the-same-name a frictionless flow and so makes this reachable in ordinary operation. It predates that PR — it applies to delete-then-recreate under *any* name — but was previously behind enough friction that nobody hit it.

**The discovery-side call at connector delete is a follow-up in that repo**; this PR only adds the capability. Nothing calls the endpoint yet, so merging it changes no behavior.

### DIFF
--- a/internal/handler/identity.go
+++ b/internal/handler/identity.go
@@ -283,6 +283,15 @@ func (a *API) registerIdentityRoutes(api huma.API) {
 		Description: "Scoped to one (origin, source_id): deactivates only still-`discovered` rows last seen before `not_seen_since`. Never touches adopted/active identities, and never another connector's rows. The offboarding half of a connector reconcile.",
 		Tags:        []string{"Identities"},
 	}, a.pruneStaleDiscoveredOp)
+
+	huma.Register(api, huma.Operation{
+		OperationID: "release-discovered-source",
+		Method:      http.MethodPost,
+		Path:        "/identities/discovered/release-source",
+		Summary:     "Unclaim a deleted connector's discovered identities so another can adopt them",
+		Description: "Scoped to one (origin, source_id): clears source_id on still-`discovered` rows so a future connector's reconcile can claim them. Called when a connector is DELETED — otherwise its rows stay bound to a source that never syncs again, and prune (which is source-scoped) can never reach them. Does NOT deactivate: adopt/dismiss remains a human decision. Adopted and dismissed rows keep their source_id as provenance.",
+		Tags:        []string{"Identities"},
+	}, a.releaseDiscoveredSourceOp)
 }
 
 type IdentitySchemaOutput struct {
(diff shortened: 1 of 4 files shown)


akhiljavelin (approved):

saucam:
Thanks — the origin-validation suggestion was right, and I verified it rather than taking it on faith: `IsExternal()` only rejects `""` and `"native"`, so `"Okta"` would have reached the `WHERE origin = ?` clause, matched nothing, and returned `released: 0`. A typo reading as *"there was nothing to release"* is exactly the kind of silent wrong answer worth closing.

Fixed in a7b7a24, with one deliberate deviation from the suggested patch:

**Placed in the service, not the handler.** ZeroID is an embedded Go module, so `IdentityService` is reachable in-process and not only over HTTP — a handler-level check wouldn't cover a direct caller. The handler already maps `ErrInvalidIdentityField` → `400` with the message, so callers get the same clear error either way, without the check living in two places. It also sits directly beside the `IsExternal` check it complements.

Added `TestDiscovered_ReleaseSourceValidation` cases for `Okta`, `OKTA`, and `okta-prod`.

**One thing I did NOT change:** `PruneStaleDiscovered` has the identical gap — it validates `IsExternal` without `ValidOrigin`, so a malformed origin there also returns `deactivated: 0` instead of a 400. Tightening it would change an existing endpoint's error surface for callers that may rely on the current behavior, which doesn't belong in this PR. Happy to do it as a follow-up if you'd like it consistent.

Re-verified: `gofmt`, `go vet`, `golangci-lint` (0 issues), `go test -race`, and the full integration suite (42.8s) — all green.

PR #296, opened 2026-08-20
chore: upgrade Go toolchain to 1.27, drop GOEXPERIMENT=jsonv2

rsharath:
## Summary
- Bumps `go.mod`/`go.work` to `go 1.27.0` across all three modules (root, `pkg/dpop`, `pkg/authjwt`) and the Docker build stage image to `golang:1.27.0-alpine3.24`.
- Removes `GOEXPERIMENT=jsonv2` from the Makefile and Docker build — Go 1.27 backs `encoding/json` with the v2 implementation by default, so the experiment flag jwx v4 needed on 1.26 is no longer required.
- Bumps `github.com/lestrrat-go/jwx/v4` to `v4.3.0` everywhere: v4.0.x/v4.2.0 hard-fail to compile on Go 1.27's finalized `encoding/json/v2` API (`jsonv2.SkipFunc` was renamed to an unexported sentinel); v4.3.0 handles both the pre- and post-1.27 stdlib shapes.

## ⚠️ Follow-up required before this is mergeable
My GitHub token lacks the `workflow` OAuth scope, so it can't push changes to `.github/workflows/*.yml`. **CI will fail on this PR as-is** (still pinned to `GO_VER: "1.26.0"` with `GOEXPERIMENT: "jsonv2"` set). Someone with `workflow` scope needs to apply the equivalent change to:
- `.github/workflows/pr-check.yml`
- `.github/workflows/release.yml`
- `.github/workflows/release-dpop.yml`
- `.github/workflows/sdk-integration.yml`

(bump `GO_VER` to `"1.27.0"`, delete the `GOEXPERIMENT: "jsonv2"` line) before this can pass and merge.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./... -short` pass on real Go 1.27.0 (root module)
- [x] Same for `pkg/dpop` and `pkg/authjwt` modules
- [x] Local `tests/integration` failure is pre-existing (no Docker daemon locally), unrelated to this change
- [ ] `highflame-docker-check` — needs CI (no local Docker daemon to verify the Dockerfile build)
- [ ] Apply the workflow YAML bump described above, then confirm CI is green


adeinega:
That was quick. 1.27 isn't available in Homebrew (yet).

JSON v2 is finally here and that's great.

One suggestion for the next PR is to switch to https://pkg.go.dev/uuid per https://go.dev/doc/go1.27#uuid.

PR #310, opened 2026-09-02
feat: remove a deleted connector's discovered identities (discovery#65)

saucam:
First half of **highflame-ai/highflame-discovery#65** — *"clean up resources on connection removal"*. This adds the ZeroID capability; the discovery-side call at connector delete is the companion PR.

## The gap

Deleting a discovery connector left every agent it found sitting in the inventory: still `discovered`, still visible, still actionable, bound to a connector that will never sync again. And since prune is scoped `WHERE source_id = ?`, **nothing could ever sweep them.**

## The change

`POST /identities/discovered/purge-source` removes the still-`discovered` rows of one `(origin, source_id)`.

### Why a DELETE and not a status flip

This is the load-bearing decision. `deactivated` is **terminal**:

```go
case IdentityStatusDeactivated:
    return target == IdentityStatusActive   // discovered is unreachable
```

…and `reconcileDiscovered` refreshes descriptive fields *"without touching lifecycle"*. So deactivating would make those agents invisible **forever** — re-adding the same connector would rediscover them, reconcile onto the dead rows, and leave them dead. That's newly reachable now that discovery#50 makes delete-then-recreate frictionless. `TestDiscovered_PurgeSourceThenReAddRediscovers` pins the working behaviour.

### What it deliberately spares

Only `discovered` rows. Anything a human has acted on — `pending` (an owner was assigned), `active`, `suspended` — has graduated past inventory and keeps both its lifecycle state and its `source_id` as provenance. Retiring a real, owned, possibly credentialed identity is a deliberate lifecycle decision, not a side effect of removing a connector. A `discovered` row has no owner and no credentials, so nothing cascades.

### Audit survives the row

`modified_by` is stamped immediately before the delete, because the audit trigger reads `OLD.modified_by` on DELETE (`012_identity_audit_trigger`, whose own comment specifies exactly this contract). Without it the audit row would attribute the removal to whoever last touched the identity. The trigger fires on `AFTER INSERT OR UPDATE OR DELETE`, so the history outlives the deleted row.

## Relationship to #292

This **supersedes `release-source` for the connector-delete path** — with the rows removed there is nothing left to unclaim. `release-source` is left in place but now has no caller. Worth deciding separately whether to retire it; I didn't want to remove a just-merged endpoint inside a PR about something else.

## Tests

Five integration tests against real Postgres:

- the acceptance criterion — the connector's agents are removed and no longer retrievable (404, not merely hidden)
- **purge → re-add → rediscovered** as a live `discovered` row owned by the new connector
- an **adopted** identity survives its connector's removal, lifecycle and provenance intact
- scoping — a purge never reaches another connector's agents
- validation — native origin, missing `source_id`, and malformed origin all rejected

```
gofmt -l .                                # clean
go vet ./...                              # clean
golangci-lint run ./...                   # 0 issues
go test ./... -race                       # green
go test -tags=integration ./tests/...     # green (48s)
```

Note: an untracked `tests/integration/scope_ceiling_probe_test.go` in my working tree (not mine, not in this PR) is missing a `//go:build integration` tag and fails both gofmt and the untagged build. I verified this change with it set aside and restored it untouched — worth someone claiming or removing it.


saucam:
⚠️ **The trivy failure shown above is stale — it is from the pre-fix commit. CI has not run since.**

**The CVE is fixed** in `8d4aec3`: `google.golang.org/grpc` v1.82.1 → v1.83.1 for **CVE-2026-84304** (HIGH). That finding was pre-existing on `main` and unrelated to this PR's feature work — `main`'s own tip is the previous bump to v1.82.1 for GHSA-hrxh-6v49-42gf, and a new advisory has since landed on that version. This PR's feature commit changes four Go files and zero dependency files.

Verified locally after the bump — worth noting `go mod tidy` carried otel `sdk/metric` + `trace` 1.43.0 → 1.44.0 and genproto along with it, so this is a wider graph move than the version bump suggests:

```
go build ./...                              ok
go vet ./...                                clean
golangci-lint run ./...                     0 issues
go test ./... -race                         green
go test -tags=integration ./tests/...       green (47.6s)
```

**GitHub Actions has not dispatched a run for this PR since `e24b07f5` (03:34 UTC).** Confirmed:

- `8d4aec3` is on the remote and listed on this PR; the external Socket checks *did* react to the push, so the event reached GitHub
- `pr-check.yml` includes `synchronize` in its trigger list
- Actions is enabled on the repo (`allowed_actions: all`)
- The most recent run in the **whole repository** is still `e24b07f5`

Tried, none of which produced a run: pushing the fix, closing + reopening this PR (`reopened` is in the trigger list), and an empty commit (`a24b949`) to force a `synchronize` event.

This looks like a GitHub-side dispatch problem rather than anything in the branch — worth a look from whoever owns CI. Once runs resume, this should go green without further changes; `a24b949` is an empty commit and can be squashed away at merge.

Separately, unrelated to this PR: an untracked `tests/integration/scope_ceiling_probe_test.go` in the repo is missing its `//go:build integration` tag, so it fails `gofmt` and compiles into the untagged build (making `go test ./...` attempt the integration suite). I set it aside for each verification run and restored it untouched — worth someone claiming or deleting it.

saucam:
✅ **Resolved — the comment above is now stale.** Both issues it describes are fixed; CI is fully green on `1c25a04`.

- **CVE-2026-84304** — grpc bumped to v1.83.1. `highflame-trivy-check` now passes in CI, not just locally.
- **Actions dispatch** — recovered on its own; the branch's push picked up a run normally.
- **`go.mod`/`go.sum` conflicts** — appeared when #304 and #307 merged into main. Resolved by rebuilding the branch on current main (cherry-pick the feature, then re-run `go get grpc@v1.83.1 && go mod tidy` against main's newer graph) rather than hand-merging, since a hand-merged module graph produces a `go.sum` matching neither side. The empty retrigger commit was dropped in the process.

Main is still on v1.82.1, so this bump is still required — not redundant with anything that landed since.

Re-verified after the rebuild, because main had moved by two PRs: build, gofmt, vet, `golangci-lint` (0 issues), unit under `-race`, the full tier-2 suite (48.6s), and the five purge tests specifically. All eleven CI checks pass.

akhiljavelin (approved):

saucam:
Both suggestions addressed in `fb6d0c3`.

**1. Duplicated WHERE clause — fixed, and it was worth more than tidiness.** The two statements must match on *exactly* the same rows: a wider delete removes identities the stamp never attributed (losing audit attribution), a narrower one leaves rows stamped but present. Two hand-maintained lists is the wrong shape for that invariant.

The suggested patch as written doesn't compile — `bun.Query` has no `Where` method. The type that does is **`bun.QueryBuilder`**, and both `*UpdateQuery` and `*DeleteQuery` expose `.QueryBuilder()`, so one `scope` function now feeds both:

```go
scope := func(qb bun.QueryBuilder) bun.QueryBuilder { ... }

stamp := db.NewUpdate().TableExpr("identities").Set("modified_by = ?", callerID)
scope(stamp.QueryBuilder())

del := db.NewDelete().TableExpr("identities")
scope(del.QueryBuilder())
```

`TestDiscovered_PurgeSourceIsScopedToItsSource` is what proves the predicate still reaches the query through the builder rather than being silently dropped — it fails if scoping is lost.

**2. Audit comment — added, with a correction to the premise.** The suggestion reads as though this path uses `SetCallerName` *instead of* an explicit stamp. It doesn't: the store **does** stamp `modified_by` explicitly (`PurgeDiscoveredSource`), and `SetCallerName` is how the value reaches it. The comment now states that two-step relationship rather than implying the caller name alone drives the trigger.

Re-verified: build, gofmt, vet, `golangci-lint` (0 issues), unit under `-race`, full tier-2 suite (49.1s), and the five purge tests specifically.

KunalJavelin (approved):

PR #319, opened 2026-09-04
fix: harden the root-mounted admin plane, finish the #318 migration, and unpin the SDK in CI

rsharath:
## Summary

Follow-up to #318 (admin API at the router root by default), addressing the post-merge review findings and settling the deployment model: **one admin path shape in SaaS and open source, differing only in who authenticates it** — AuthN's `AdminAuth` hook in SaaS, the network layer (or an `AdminAuth` hook when embedding) for self-hosters. The SDKs work against any deployment with just a base URL; no client options, no env vars, no pins.

## Security / deployment guidance

- **Startup warning** on a root-mounted deployment that has no `AdminAuth` hook installed (checked at `Start`, so embedders like highflame-authn — which set the hook — never see it). Says what to do; names the pre-#318 network-rule break.
- **CIBA approve/deny are secured by guidance, mirroring SaaS.** `POST /oauth2/bc-authorize/{id}/approve|deny` take the approving human's `subject_id` in the body, so they need a user-auth gateway in production regardless of path — exactly how SaaS runs them (AuthN resolves the approver from the signed-in session). The README's Admin section now states that they live inside `/oauth2/*` and are **not** covered by a public-prefix allowlist, gives the edge rule to deny them ahead of the `/oauth2/` allow, and notes that deployments not using CIBA still need it (the endpoints are registered regardless; the default policy permits the CIBA grant). The startup log and the warning no longer describe `/oauth2/*` as uniformly public. The paths themselves are unchanged — no cross-repo lockstep with authn/studio.
- README Admin section rewritten around the SaaS-mirroring model and the `/api/v1 → root` migration note.

## Correctness

- **OpenAPI**: admin spec/docs/schemas at `{mount}/admin/*` (`NewHumaAPIAdmin`) — the ~70 admin operations are documented again and response `$schema` links resolve (verified live: `/openapi.json` 13 public paths, `/admin/openapi.json` 51 admin paths, `$schema` → 200). Specless instances drop the link transformer instead of stamping dead links; the agent-auth group is specless everywhere, fixing the pre-existing prefixed-mount spec shadow behind its 401 middleware (and the same latent shadow in highflame-authn).
- **`zeroid token revoke`** sends RFC 7009 client authentication (`--client-id`/`--client-secret` or `ZID_CLIENT_ID`/`ZID_CLIENT_SECRET`) via direct fetch, since the SDK can't carry it; missing credentials fail fast with guidance toward `agents deactivate`. Previously guaranteed to 401 in production.
- **Migration sweep**: remaining `/api/v1` call sites — `examples/a2a_org_hierarchy/provision.py`, both agent `SKILL.md` files, README endpoint table, `cli/README.md` admin-prefix guidance.
- **Quickstart**: RFC 7009 claim states the dev-mode carve-out instead of asserting a guarantee the compose deployment doesn't enforce.

## CI

- **SDKs installed unpinned** (`pip install --upgrade highflame`, `@highflame/sdk@latest`) in every job, so an SDK release that stops speaking to a stock ZeroID fails CI instead of hiding behind a pin — the exact failure mode that started this series.
- **SDK integration workflow runs nightly** (`17 6 * * *`) plus `workflow_dispatch`, because PR-triggered runs only notice a new SDK release when someone opens a PR.

## Tests

- TS smoke restores `/oauth2/token/revoke` coverage with a raw authenticated revoke (TS SDK gap noted inline).
- CLI config tests standardize on `ZID_CONFIG_DIR` isolation (homedir mock and `HOME_CONFIG` spreads removed); the init tests were writing into the developer's real `~/.config/zeroid`.

## Deliberately not done

- CIBA path move (see above) — closed highflame-authn#193 and highflame-studio#1555.
- CLI per-invocation admin-token mint (efficiency finding) — caching a bearer on disk deserves its own discussion.

## Verification

`go vet`; 497 unit + 680 integration tests with `-race`; 112/112 CLI tests; TS SDK smoke 8/8; live checks against a fresh compose build (specs, `$schema` 200, startup warning, public `bc-authorize` routing). CI green with unpinned SDKs.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -189,7 +189,7 @@ jobs:
         shell: bash
         run: |-
           pip install --quiet "testcontainers[postgres]" "PyJWT>=2.8" cryptography pytest
-          pip install --quiet "highflame==0.3.23"
+          pip install --quiet --upgrade highflame
 
       - name: Run Python SDK smoke tests
         run: pytest tests/sdk/test_sdk_smoke.py -v --tb=short -x
@@ -200,7 +200,7 @@ jobs:
         run: |-
           npm init -y
           npm install --save-dev vitest typescript @testcontainers/postgresql
-          npm install @highflame/sdk@0.3.23
+          npm install @highflame/sdk@latest
 
       - name: Run TypeScript SDK smoke tests
         working-directory: tests/sdk
@@ -259,7 +259,7 @@ jobs:
 
       - name: Install notebook execution deps
         shell: bash
-        run: pip install --quiet pytest nbmake ipykernel "highflame==0.3.23" cryptography "PyJWT>=2.8"
+        run: pip install --quiet --upgrade pytest nbmake ipykernel highflame cryptography "PyJWT>=2.8"
 
       - name: Execute the quickstart notebook against the live server
         shell: bash
--- a/.github/workflows/sdk-integration.yml
+++ b/.github/workflows/sdk-integration.yml
@@ -11,9 +11,16 @@ on:
   merge_group:
     types:
       - checks_requested
+  # The SDKs are installed unpinned (latest release) so an SDK version that
+  # stops speaking to a stock ZeroID fails here instead of hiding behind a
+  # pin. PR-triggered runs only notice when someone opens a PR, so also run
+  # nightly against main.
+  schedule:
+    - cron: "17 6 * * *"
+  workflow_dispatch:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 
 env:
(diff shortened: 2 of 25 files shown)


rsharath:
Superseded by #321, which reverts #318's default-prefix flip (back to `/api/v1`) and carries forward the pieces of this PR that are correct under either mount shape: the CLI `token revoke` client-auth fix, CLI test isolation, the OpenAPI shadow fixes (agent-auth specless everywhere, `/admin/*` spec for root mounts), the specless `$schema` cleanup, and the gated startup warning. The CIBA approve/deny path move is dropped — it only existed to plug an exposure created by the root mount, which no longer exists under a prefix default.

rsharath:
Reopening — decision reversed after discussion: we're keeping one admin path shape everywhere (root mount, as the SDKs speak it), with SaaS/AuthN authenticating via AdminAuth and self-hosters at the network layer. This PR's hardening — moving CIBA approve/deny out of `/oauth2/*` so the public surface is clean for allowlists, the `/admin/*` OpenAPI fix, the gated startup warning, the CLI revoke fix, the `/api/v1` doc sweep — is what makes that shape safe. #321 (the revert) is closed instead. Adding the CI change (install latest SDK unpinned + nightly run) here too.

jalbrethsen-highflame (approved):

PR #321, opened 2026-09-04
fix: restore /api/v1 as the default admin mount; keep the CLI/OpenAPI/auth fixes

rsharath:
## Summary

Reverts the default admin-mount flip from #318 and supersedes #319. The default goes back to `server.admin_path_prefix: "/api/v1"`.

**Why revert.** ZeroID's admin plane has no built-in auth by design — self-hosted deployments protect it at the network layer, and a distinct prefix is what keeps that simple (`deny /api/v1/*`). Mounting admin at the root (a) interleaved unauthenticated admin routes with the public `/oauth2/*` surface — which is how the CIBA approve/deny endpoints ended up inside a namespace every doc calls public, (b) silently broke every existing self-hosted deployment's network rules on upgrade, and (c) forced a three-repo lockstep (authn, studio) to chase the resulting path move. None of that exists under a prefix. Embedders that gate the admin group themselves — highflame-authn and the SaaS deployment — already set `""` explicitly and are unaffected either way.

**What about the SDK mismatch that motivated #318?** SDK ≥ 0.3.23 addresses admin routes at the root, so it 404s against stock OSS ZeroID. That belongs on the SDK side: a client-side admin base-path option (highflame-sdk PR to follow), with metadata-advertised auto-discovery as the phase-two zero-config path. Until it ships, CI pins return to 0.3.17 — the state before #318, which worked.

## Kept (correct under either mount shape)

- **CLI on `@highflame/sdk` 0.3.23** — root paths are the SaaS/AuthN shape the CLI targets by default (`api.zeroid.io`); self-hosted users pass `--admin-prefix /api/v1`. `tokens.issueApiKey` migration, mint-handler test mocks, `cli/README` guidance corrected.
- **`zeroid token revoke`** sends RFC 7009 client authentication (the SDK can't carry it); missing credentials fail fast with guidance toward `agents deactivate`.
- **CLI test isolation** via `ZID_CONFIG_DIR` — the init tests were writing profiles into the developer's real `~/.config/zeroid`.
- **OpenAPI**: the agent-auth Huma group is specless everywhere — its spec route sat behind the 401 middleware and shadowed `{prefix}/openapi.json` (verified live: `/api/v1/openapi.json` now serves all 51 admin paths; before it returned 401). Root-mounted admin groups serve their spec under `/admin/*` so they never shadow the public `/openapi.json` — this also fixes the same latent shadow in highflame-authn. Specless instances drop the `$schema` link transformer instead of stamping dead links.
- **Startup warning** when a deployment chooses `""` without installing an AdminAuth hook.

## Verification

`go vet`; 497 unit + 680 integration tests with `-race`; 112/112 CLI tests; Python SDK smoke 8/8 on 0.3.17 (testcontainers); quickstart executed against a fresh compose build; live: `/api/v1/identities` → 200, `/identities` → 404, admin spec at `/api/v1/openapi.json`, no warning on a prefixed deployment.

## Related

- Supersedes #319 (closing). highflame-authn#193 and highflame-studio#1555 (the CIBA path followers) are closed — not needed under a prefix default.
- #320 (default-policy CIBA migration) is independent and unaffected.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -189,7 +189,7 @@ jobs:
         shell: bash
         run: |-
           pip install --quiet "testcontainers[postgres]" "PyJWT>=2.8" cryptography pytest
-          pip install --quiet "highflame==0.3.23"
+          pip install --quiet "highflame==0.3.17"
 
       - name: Run Python SDK smoke tests
         run: pytest tests/sdk/test_sdk_smoke.py -v --tb=short -x
@@ -200,7 +200,7 @@ jobs:
         run: |-
           npm init -y
           npm install --save-dev vitest typescript @testcontainers/postgresql
-          npm install @highflame/sdk@0.3.23
+          npm install @highflame/sdk@0.3.17
 
       - name: Run TypeScript SDK smoke tests
         working-directory: tests/sdk
@@ -259,7 +259,7 @@ jobs:
 
       - name: Install notebook execution deps
         shell: bash
-        run: pip install --quiet pytest nbmake ipykernel "highflame==0.3.23" cryptography "PyJWT>=2.8"
+        run: pip install --quiet pytest nbmake ipykernel "highflame==0.3.17" cryptography "PyJWT>=2.8"
 
       - name: Execute the quickstart notebook against the live server
         shell: bash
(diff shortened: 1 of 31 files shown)


rsharath:
Closing without merge — decision reversed after discussion. We're keeping one admin path shape in SaaS and open source (root mount, which the SDKs speak natively), differing only in who authenticates the admin plane: AuthN's AdminAuth hook in SaaS, the network layer for self-hosters. #319 (reopened) carries the hardening that makes the root mount safe. The pieces of this PR that were mount-independent already live in #319.

Issue #327, opened 2026-09-06
resource at /oauth2/authorize is ignored — a client that sends it only there gets an unbound token


saucam:
Follow-up from the RFC 8707 `resource` parameter work (#258, `CAP-IDN-026`). Found in adversarial review; not a live exploit, but a silent-unbound-token shape that survives only because of what MCP clients happen to do.

## What happens today

`resource` establishes a binding **only** at `POST /oauth2/token`. `GET /oauth2/authorize` is a plain chi handler that reads its named parameters and ignores everything else (RFC 6749 §3.1's ignore-unrecognized posture), and the binding is not persisted on the authorization code.

So:

| Client sends `resource` at | Result |
|---|---|
| authorize **and** token | bound correctly |
| token only | bound correctly |
| **authorize only** | **200 with an UNBOUND token, no error anywhere** |

The third row is the forbidden shape: the client believes it holds a token narrowed to one MCP server; Shield's `checkResourceBinding` short-circuits to allow when the `resource` claim is absent, so the token is honoured at every MCP server in the tenant.

## Why it hasn't bitten

The MCP authorization profile makes `resource` a **MUST** on both the authorization request and the token request, so a conformant MCP client lands in row 1. A client following **RFC 8707 §2 alone** can reasonably send it only at the authorization request — §2 has the AS bind at authorization time, with the code carrying the grant — and lands in row 3.

## Why "just reject it at /oauth2/authorize" is wrong

That would 400 every conformant MCP client at the first step of the flow, since the MCP profile requires the parameter there. Rejecting is not available.

## The actual fix

1. `/oauth2/authorize` validates `resource` (same RFC 8707 §2 syntax rules as the token endpoint) and **persists the authorized set on the authorization-code row** — a migration adding a `resource` column, the same shape migration `039` established for `audience` on `refresh_tokens`.
2. `authorizationCode` binds from the row when the token request omits `resource`.
3. When the token request **also** sends `resource`, each value must be a **subset** of the row's set, else `invalid_target` — the same narrows-only rule the ID-JAG path uses (`narrowResourcesTo`). The client selects; it can never add.

That makes all three rows correct and matches what RFC 8707 §2 actually describes.

## Interaction to keep in mind

[...]

PR #329, opened 2026-09-06
feat: RFC 8707 resource request parameter on the token endpoint (CAP-IDN-026)

### LINKED ISSUE #258: RFC 8707 resource parameter on the token endpoint — the minting half of INV-IDN-006
Follow-up to highflame-ai/highflame-shield#366. Spec context: highflame-ai/highflame-architecture#258.

## Why

#366 fixed the *enforcement* half of INV-IDN-006 — Shield now gates on an explicit RFC 8707 `resource` claim instead of inferring a binding from `aud` (zeroid#257 stamps it, highflame-shield#368 reads it).

But the *minting* half barely exists. Today the only path that produces a resource-bound token is the ID-JAG grant, and ID-JAG is not configured in prod. **So INV-IDN-006 is now correct and dormant: it enforces nothing.**

That is the right place to have landed the bug fix, but it is not a finished capability.

## What is missing

1. **No RFC 8707 `resource` parameter on `/oauth2/token`.** There is no way for an ordinary grant (`client_credentials`, `api_key`, `token_exchange`) to request a token bound to a specific MCP server without an enterprise IdP in the loop.
2. **Firehog does not advertise RFC 9728 protected-resource identifiers** (`<origin>/mcp/<slug>`) per MCP server. Shield's `resourceAuthorizesServer` has a slug-tail matching arm built for exactly that shape, and nothing currently produces it — callers would have to pass bare server names or exact URLs.
3. **No consumer.** Cerberus does not request bound tokens today. Deciding *who* asks for one, and against what, is the actual design question — the parameter plumbing is the easy part.

## The trap to design around

ZeroID already overloads `aud` for **audience profiles** (`codeoid`, `agent-sandbox` — `internal/service/oauth.go`, `defaultAudienceScopeProfiles`). Those are scope-profile labels, not resources.

A `resource` request parameter arriving alongside the existing `audience` parameter needs **defined precedence**. Getting that wrong re-creates exactly the class of bug #366 was: two different meanings sharing one channel with no discriminator. Worth settling before any code:

- Are `audience` and `resource` mutually exclusive on a single request?
- If both are supplied, which wins — or is it an `invalid_request`?
- Does a profiled audience token ever also carry a resource binding?

## Scope note

This needs its own `CAP-IDN-*` entry rather than widening INV-IDN-006 — it is a new platform capability (the ability to mint resource-bound tokens outside ID-JAG), not a clarification of an existing invariant. `/feature-prep` before implementation.

saucam:
Closes #258. Implements `CAP-IDN-026`.

## Why

`INV-IDN-006` enforces a binding that almost nothing produces. The ID-JAG grant was the only path that stamped `resource`, and ID-JAG is not configured in prod — so the invariant was **correct and dormant**. This is the minting half: RFC 8707 §2 `resource` as a **request parameter** on `POST /oauth2/token`, available to ordinary grants with no enterprise IdP in the loop.

## What it does

| Grant | Behaviour |
|---|---|
| `jwt-bearer` (ID-JAG) | **SELECTS** — each requested value must appear in the assertion's own `resource` claim, else `invalid_target`. The client picks which MCP server; today the claim picked for it. |
| `client_credentials`, `api_key`, `token-exchange` (all 3 forks), `authorization_code` | **BINDS** — no upstream authorized set to narrow against, so the request restricts the token that would have been minted anyway. |
| `refresh_token`, CIBA, unknown/custom grants | **REFUSED** with `invalid_target`. Accepted-and-ignored is the one outcome that must never happen. |

Two claims are stamped, because they answer different questions: `aud` for RFC 8707 §2 conformance, and `resource` — the discriminator Shield's `INV-IDN-006` enforcement keys on. It cannot read `aud` for that: ZeroID stamps `aud` on every token it issues, defaulting to the issuer URL for JWT-SVID §3, so a non-empty `aud` says nothing about whether a binding exists. Reading it as if it did denied every MCP-targeted request in prod (highflame-shield#366).

## Three decisions worth reviewing explicitly

**A resource-bound request receives no refresh token.** `authorization_code` is the only bindable grant that mints one, and the refresh path re-mints from state on the refresh-token row — where the binding is not, because it lives in `CustomClaims`, which rotation does not carry. A rotated token would come back **unbound, silently**: bind to server X, refresh, use anywhere, no error at any step, and enforcement goes quiet because the discriminator claim is simply gone. Suppressing fails closed and needs no migration. Carrying it on the row (a `resource` column, mirroring migration `039`'s `audience`) is the alternative if long-lived resource-bound sessions ever have a real use case.

**`audience` and `resource` are mutually exclusive** — both on one request is `invalid_request`. `audience` names a server-defined scope profile (it widens); `resource` names a resource (it narrows). Two meanings sharing one channel with no discriminator is precisely the bug `INV-IDN-006` was narrowed to fix; this refuses to re-create it one parameter over.

**`aud` is not an authorization signal.** The parameter also sets `aud`, so on the ordinary grants the **caller chooses** that claim. "Only narrows" holds for the presence-gated `resource` claim; it does not extend to `aud`, which is an accept/reject gate. Safe because nothing on the platform authorizes on `aud` — verified across Shield, Observatory, Cerberus, Discovery and AuthN, all of which build their verifier with issuer + JWKS and leave the audience check unset. Recorded in highflame-ai/highflame-architecture#330, and `pkg/authjwt`'s `VerifierConfig.Audience` godoc is corrected here — it previously said *"Use your service identifier (e.g. `my-mcp-server`)"*, which reads as proof of intended recipient and is not.

## A spec conflict this had to resolve

RFC 8707 §2 makes `resource` **repeatable**; `oauthFormCompatMiddleware` enforced RFC 6749 §3.1's blanket "MUST NOT be included more than once". A conformant multi-resource form request was rejected with `duplicate OAuth parameter: resource` — which reads to an interop tester as "resource unsupported", not "we disagree about §3.1". Added a carve-out scoped to `/oauth2/token` and to that one field; a duplicated `client_id` is still rejected, and that is tested.

## Review found a real hole, and it is fixed here

`token-exchange` forks three ways and the grant type as a whole passes the gate. Only two forks bound. `externalIDTokenExchange` returned **200 with an unbound token** — the caller believes it is narrowed to one MCP server, Shield honours it at all of them. Structurally identical to the `jwt-bearer` ID-JAG/NHI split that *was* guarded; I saw that split and missed this one (`fdab65f`).

The bigger finding was a **test gap**: deleting the ID-JAG narrowing call left the entire suite green. Coverage is now per **issuance site** rather than per grant-type string — that is the unit that can independently forget to bind — and all six are mutation-checked. Before the fix, 2 of 5 were killed.

Adversarial review then found two more: a **userinfo component** was accepted into a resource indicator, putting a credential verbatim into a signed JWT, the credential row, the observed-resource inventory and the logs (`a231a02`); and malformed values were answered with a huma **422** rather than the RFC 6749 §5.2 error object, which had also made the service's own parse-error branch unreachable over HTTP (`70d45f2`).

## Verification

- **Exercised against a running service**, not just harnesses: an isolated ZeroID (own binary, own Postgres) — 21/21 on the ordinary grants, and 13/13 on ID-JAG narrowing against a real ES256 stub OP registered as an external issuer, including that a rejected `resource` does **not** burn the single-use `jti`.
- **Cross-repo shape check**: the exact identifier this mints, fed to Shield's `resourceAuthorizesServer` — authorizes the bound server, denies a different server behind the same gateway, denies a prefix lookalike (`github-admin` vs `github`), denies same-slug-different-origin.
- Both CI gates green after a rebase onto current `main`: 8 unit packages under `-race`, integration at ~52s.

## Lands in lockstep with the regression PR

`highflame-regression` carries `test_row_resource_as_request_parameter` as an **`xfail(strict=True)`**. The moment a build with this change reaches that environment the test XPASSes, and `strict` turns that into a suite failure. The companion PR removes the marker and adds `@pytest.mark.capability("CAP-IDN-026")`, which is what flips the catalog entry from `planned` to `implemented`.

## Known gaps, filed rather than hidden

- **#327** — `/oauth2/authorize` ignores `resource`. A client that sends it *only* there gets an unbound token. Safe today because the MCP profile requires it at both endpoints; a client following RFC 8707 §2 alone is not. The real fix persists the authorized set on the auth-code row and cross-checks at redemption — a migration, out of #258's scope. Documented on the field's OpenAPI description in the meantime.
- **highflame-firehog#628** — `oauth.expected_audience` is a single global `Option<String>` that would reject every resource-bound token if set, and can never be correct against per-slug RFC 9728 identifiers. Unset everywhere today.
- Shield's `sameOrigin` clause vs Firehog's relay path (`server_url` is the upstream URL, not the gateway origin), and Firehog's global PRM advertising a pathless `<origin>` that `resourceSlug` cannot match. Both follow-ups.
- **Issue #258 item 3** (who actually requests bound tokens) is unanswered by design — the parameter is a prerequisite either way. Item 2 (Firehog per-server RFC 9728 metadata) was **already shipped**; no Firehog work was needed.

## Not claimed

`make test` cannot pass and could not before this change — it runs `./...` at `-timeout=120s` against an integration suite that takes ~420s under `-race`. Verified against the two commands `pr-check.yml` actually runs instead. The Makefile target is stale and unrelated; left alone.

### DIFF
--- a/internal/handler/oauth.go
+++ b/internal/handler/oauth.go
@@ -94,7 +94,23 @@ type TokenInput struct {
 		// resource(s) the minted token is bound to (CAP-IDN-026). Accepts a
 		// single URI string or an array of them; a form-encoded request may
 		// repeat the parameter. Mutually exclusive with Audience.
-		Resource resourceParam `json:"resource,omitempty" doc:"RFC 8707 resource indicator(s) to bind the token to — absolute URI(s), no fragment"`
+		//
+		// THE TOKEN REQUEST IS THE ONLY PLACE A BINDING IS ESTABLISHED.
+		// /oauth2/authorize accepts and ignores `resource` (RFC 6749 §3.1's
+		// ignore-unrecognized posture), and the binding is not carried on the
+		// authorization code — so a client that sends it only at the
+		// authorization request receives an UNBOUND token, with no error at any
+		// step. The MCP authorization profile requires `resource` on both the
+		// authorization and the token request, so a conformant MCP client is
+		// bound correctly; a client following RFC 8707 §2 alone, which permits
+		// the token request to rely on the code's binding, is not.
+		// Tracked for a follow-up that persists the authorized set on the code
+		// row and cross-checks it at redemption.
+		//
+		// Note also that a request carrying `resource` is issued NO refresh
+		// token: the binding is not carried across rotation, so a refresh would
+		// silently unbind the token.
+		Resource resourceParam `json:"resource,omitempty" doc:"RFC 8707 resource indicator(s) to bind the token to — absolute URI(s), no fragment. Must be sent on THIS request: /oauth2/authorize ignores it and the binding is not carried on the authorization code. A resource-bound request receives no refresh_token."`
 		// authorization_code grant fields:
 		Code         string `json:"code,omitempty" doc:"Authorization code JWT"`
 		CodeVerifier string `json:"code_verifier,omitempty" doc:"PKCE S256 code verifier"`
--- a/internal/handler/oauth.go
+++ b/internal/handler/oauth.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
@@ -89,6 +90,11 @@ type TokenInput struct {
 		// external principal can self-rotate its session at /oauth2/token. Honoured
 		// only on that path AND only for a profiled Audience; ignored otherwise.
 		IssueRefreshToken bool `json:"issue_refresh_token,omitempty" doc:"Also mint a rotating refresh token (trusted external-principal exchange with a profiled audience only)"`
+		// Resource is the RFC 8707 §2 resource indicator: the protected
+		// resource(s) the minted token is bound to (CAP-IDN-026). Accepts a
+		// single URI string or an array of them; a form-encoded request may
+		// repeat the parameter. Mutually exclusive with Audience.
+		Resource resourceParam `json:"resource,omitempty" doc:"RFC 8707 resource indicator(s) to bind the token to — absolute URI(s), no fragment"`
 		// authorization_code grant fields:
 		Code         string `json:"code,omitempty" doc:"Authorization code JWT"`
 		CodeVerifier string `json:"code_verifier,omitempty" doc:"PKCE S256 code verifier"`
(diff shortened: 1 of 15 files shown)


saucam:
**Lockstep merge partner:** highflame-ai/highflame-regression#224

That PR removes the `xfail(strict=True)` on `test_row_resource_as_request_parameter`. Merging this one without it turns the regression suite **red** — the test starts passing, and `strict=True` turns a pass into a failure.

Merge order that works: either together, or #224 first (it is inert until a build with this change reaches the regression environment).

saucam:
Sharpening the lockstep note above, because the detection story is worse than "the suite goes red" implies.

`highflame-regression`'s **PR gate never runs pytest** — `pr-check.yml` is env + commit-check + trivy only. The MCP interop rows run in the **nightly kind workflow**. So:

- Merging this PR without highflame-ai/highflame-regression#224 will **not** show up as a red check on any PR.
- The first signal is a **nightly failure**, hours later, on a suite nobody is watching at merge time.
- The cause will look unrelated to this change: a strict-xfail row that started *passing*.

That makes merge order a real decision rather than a formality. #224 is inert until a build containing this change reaches the regression environment, so **merging #224 first is safe at any time** and removes the window entirely.

Also worth stating plainly for anyone weighing the evidence here: green CI on #224 says nothing about whether `test_row_resource_as_request_parameter` actually passes now. Neither PR's checks execute it. What was verified is the *behaviour* it asserts — ID-JAG narrowing exercised against a running ZeroID with a real ES256 stub OP registered as an external issuer, including that a rejected `resource` does not burn the single-use `jti`. That is strong evidence, but it is not the test having been run.

saucam:
pr-shepherd: CI is green at `1204437` — all 11 checks SUCCESS, no auto-fixes were needed.

<details>
<summary>Checks (head <code>1204437159566265cbf531cdc20d842545890ebc</code>)</summary>

| check | result |
| --- | --- |
| highflame-build-check | pass |
| highflame-lint-check | pass |
| highflame-integration-test | pass (10m33s) |
| highflame-cli-check | pass |
| highflame-docker-check | pass |
| highflame-notebook-check | pass |
| highflame-trivy-check | pass |
| Python SDK smoke tests | pass |
| TypeScript SDK smoke tests | pass |
| Socket Security: Project Report | pass |
| Socket Security: Pull Request Alerts | pass |

Within `highflame-integration-test`, both Go gates passed independently: `Run Go unit tests (race detector)` and `Run Go integration tests in tests/`.
</details>

Nothing mechanical to repair: all 12 commit subjects already use the allowed `feat:`/`fix:` prefixes, `gofmt -l` is clean, `go vet ./...` is clean at the tip, no go.sum drift, and Trivy reported no findings to bump. Branch is 12 ahead / 0 behind `main`, so no rebase was performed and nothing was force-pushed — the commits reviewed here are exactly what CI tested.

**Not merging.** Per the lockstep note above, merge order is a human decision and highflame-ai/highflame-regression#224 should land first.

rsharath (approved):

PR #330, opened 2026-09-07
fix: emit the RFC 9068 `client_id` claim on OAuth-issued access tokens (#325)

### LINKED ISSUE #325: CIMD client_id does not survive the code→token exchange, leaving resource servers unable to identify the client
## Summary

The authorization code carries the CIMD `client_id`, but the access token
exchanged for it does not. A resource server therefore cannot tell which client
a call came from — which removes most of the operational value of accepting
unregistered clients in the first place.

## Observed

Full CIMD authorization_code flow on dev1, real browser consent leg.

**Authorization code** (decoded) — client identity present:

```json
{
  "cid": "https://studio-dev.highflame.dev/.well-known/oauth-client/highflame-reference.json",
  "cc":  "<code_challenge>",
  "ruri":"http://127.0.0.1:54321/oauth/callback",
  "aid": "428956146418",
  "pid": "125bff3b-73df-44c2-ab13-1f34e6563108",
  "uid": "user_3FX8…",
  "sub": "auth-code"
}
```

**Access token** from `/oauth2/token` — client identity gone:

```
iss        = https://auth-dev.highflame.dev
aud        = ["https://auth-dev.highflame.dev"]
sub        = user_3FX8…
scope      = account:read gateway:read mcp:tools
account_id = 428956146418
project_id = 125bff3b-…
client_id  = (absent — checked both `client_id` and `cid`)
```

## Why it matters

With DCR, every client has a registration row, so `client_id` is recoverable
server-side even if the token omits it. CIMD deliberately has **no registration
row** — the document URL *is* the identity. If it is not in the token, the
identity does not survive the exchange and nothing downstream can recover it.

Concretely, this blocks:

- **Attribution** — Firehog spans and Observatory events can name the user and
  tenant but not the client. "Which MCP client made this call" is unanswerable.
- **Per-client policy** — no Cedar principal to write against, so every CIMD
  client is indistinguishable at the resource server.
- **Revocation of a bad client** — with no client identity in the token, the
  only lever is revoking users or the whole issuer.

This compounds with highflame-firehog#627 (granted scope not enforced): between
them, a resource server can distinguish neither *who* the client is nor *what*
it was granted.

## Suggested fix

`IssueRequest` already supports `CustomClaims` (used by the ID-JAG path,
`internal/service/oauth_id_jag.go:333`), so the authorization_code path can
carry `cid` through from the code into the access token without new plumbing.

Worth deciding deliberately:

- claim name — `client_id` is the RFC 9068 §2.2 registered name for this and is
  what a resource server will look for; `cid` is what the code already uses
- whether it belongs on refresh tokens too, so the identity survives renewal
- length — a CIMD `client_id` is a URL, and Studio already caps it at 255 chars
  for VARCHAR(255) columns (`packages/auth/src/cli-auth/constants.ts`); token
  bloat is worth a glance but not a blocker

Found while proving the dev1 CIMD flow end to end — the flow itself works.

saucam:
Closes #325 — though **not as that issue originally described it**, and the correction is worth reading before the diff.

## The issue was wrong about the mechanism

It claimed the CIMD `client_id` "does not survive the code→token exchange". **It does** — as `application_id`:

```go
// internal/service/oauth.go — authorization_code
ApplicationID: authCode.ClientID,

// internal/service/credential.go:454
if req.ApplicationID != "" {
    _ = token.Set("application_id", req.ApplicationID)
}
```

My original evidence was a token decode that checked `client_id` and `cid` and found neither. That was true; the conclusion wasn't. I never checked `application_id`. [Corrected on the issue.](https://github.com/highflame-ai/zeroid/issues/325#issuecomment-5565855051)

## What was actually wrong

**The claim name.** RFC 9068 §2.2 registers `client_id` for exactly this, and it's what a resource server reads to attribute a call. `application_id` is a Highflame name whose documented meaning is broader — *"the optional application scope, set when an API key is linked to an application"* — so a partner's MCP server has no reason to look there. That matters directly for the interop work.

For a CIMD client, the client_id is the metadata-document URL — the whole of that client's identity, since there's no registration row.

## The change

A dedicated `IssueRequest.ClientID`, emitted as `client_id`, on the authorization_code and refresh_token paths.

**Deliberately not done by widening `ApplicationID`.** The two mean different things, and conflating them would stamp `client_id` onto api_key tokens that have no OAuth client at all. `application_id` keeps its existing value — this adds a claim rather than renaming one, so existing consumers are untouched.

## Why the refresh path is in here

A refreshed access token is minted by a **different** `IssueCredential` call in `refreshToken`. Wiring only the authorization_code path would give a token that identifies its client for an hour and then stops — the gap reappearing at renewal, where nobody would think to look.

Both paths are mutation-tested independently:

| Mutant | Tests failed |
|---|---|
| remove authorization_code wiring | 2 |
| remove refresh wiring | **1** |

That the refresh mutant fails exactly one is what proves that test isn't redundant with the first.

## Not in scope

- `client_credentials` and the ID-JAG path also have a client, but they set `ApplicationID` from a genuine application scope rather than a client_id, so they need their own decision about where `client_id` comes from.
- **Downstream, nothing consumes this yet.** Firehog parses `application_id` into `HighflameClaims` and never reads it — `grep -rn "claims.application_id"` over `src/` returns nothing. Same shape as highflame-firehog#627, where the granted scope was parsed, forwarded and audited but never compared against anything. Consuming `client_id` for attribution is follow-up work there.

## Verification

```
go build ./...   clean
go vet ./...     clean
go test ./...    9 packages ok, 0 FAIL (incl. tests/integration, 51s)
```

Note CI also runs the suite under `-race`, which my local run didn't — flagging it since that gap has bitten me before on this repo.


saucam:
## Self-review — found one real bug, in the change itself

Reviewing my own PR has an obvious bias problem, so I went looking for reasons it was wrong rather than confirmation it was right. One finding, and it matters more than the original issue did.

### The claim I added was forgeable (fixed in `fe90812`)

`client_id` was not in `reservedClaims`, and the ordering makes that exploitable:

```
req.AdditionalClaims  (caller-supplied)
  → filtered by reservedClaims          ← client_id was not in the map
  → customClaims
  → IssueRequest.CustomClaims
  → applied in IssueCredential at credential.go:489
```

The dedicated set is at `credential.go:479` — **before** that loop — so a CustomClaims entry wins. An `additional_claims: {"client_id": "https://victim/doc.json"}` on the `ExternalPrincipalExchange` path lands in the token verbatim.

That path sits behind `TrustedServiceValidator`, so the caller must be an authorised trusted service. That lowers severity but doesn't dismiss it: `reservedClaims` exists precisely so those callers *still* can't override standard claims — which is why `cnf`, `role` and `privilege_scope` are already in it.

**A forgeable attribution claim is worse than no attribution claim**, because a consumer that trusts it has no way to tell. And for a CIMD client it's total impersonation — the metadata-document URL is the entire identity, since there's no registration row behind it.

Fixed by reserving `client_id`, with a tripwire test in the house style of `TestResourceIsReserved`. Mutation-checked: removing the reservation fails 4 assertions.

### Things I checked that came back clean

- **Refresh binding** — `req.ClientID` on the refresh path is validated against the refresh token's own client before issuance, so a client can't refresh another's token and stamp its own `client_id`.
- **Double-set** — nothing else writes a `client_id` claim, so there's no last-writer ambiguity beyond the CustomClaims case above.
- **ES256 tokens** — the claim sits in the RS256 block alongside `application_id`; both authorization_code and refresh set `UseRS256: true`, so both are covered.

### Left deliberately undone

- **`application_id` is still unreserved.** Equally forgeable, and pre-existing. Reserving it now could break a trusted-service caller legitimately setting it, so it wants its own change. Recorded as an executable decision in `TestReservedClaimsCoversTheAttributionClaims`, which fails if someone reserves it without saying why.
- **The ID-JAG path** can write `client_id` via `cfg.PropagateClaims` — deployment config, not caller input, so operator intent rather than a vector.

### Reviewer note

I'd value a second pair of eyes specifically on whether `ExternalPrincipalExchange` callers in the wild set `client_id` in `additional_claims` today. If any do, reserving it turns a silently-honoured claim into a silently-dropped one. I found no such caller in this repo, but I can't see deployment config from here.

`go vet` clean, full suite green (9 packages, including `tests/integration`).

akhiljavelin (approved):

saucam:
## Deep audit — one finding, folded in as `dad9edd`

Rebased onto your `main` merge (`fae18a8`), preserved rather than force-pushed over.

### Finding: the refresh path read `client_id` from the request, not the grant

```go
applicationID := oldToken.ClientID   // line 2417 — the stored grant
...
ClientID: req.ClientID,              // mine — the request redeeming it
```

Two claims meant to name the same client, read from two different places.

They're equal in practice — I verified empirically that a cross-client refresh is rejected: presenting client A's refresh token as client B answers `400 invalid_grant`. **But the only `client_id` check I could locate is inside the `audienceRefreshToken != nil` branch**, so for an ordinary refresh token — the normal CIMD/MCP case — the equality is upheld somewhere I couldn't point at.

Deriving an attribution claim from a request field whose validation I can't locate is the wrong trade. Both now come from `oldToken.ClientID`, so they're provably the same value and neither depends on that check.

The refresh test now also asserts `client_id == application_id`. A token asserting two *different* clients would be worse than one asserting none.

### Checked, no change needed

- **Reserving `client_id` doesn't break external-IdP claim mapping.** `propagate_claims` is restricted to `auth_time/acr/amr` by `ExternalIssuerConfig.Validate`, and that loop already filters `reservedClaims` deliberately.
- **The ID-JAG propagate loop does *not* filter `reservedClaims`** — but the same config validation means `client_id` can't be named there. Worth knowing the guarantee is held one file away.
- **No other writer** of a `client_id` claim, so no last-writer ambiguity beyond the CustomClaims ordering already fixed in `fe90812`.
- **ES256 tokens** — the claim sits in the RS256 block; both affected paths set `UseRS256`.

### Two things I want a reviewer on, because I can't settle them

1. **Where is a non-audience cross-client refresh rejected?** It demonstrably is, but I couldn't find the code. If it turns out *not* to be enforced in some path, that's an RFC 6749 §10.4 issue well beyond this PR — and worth knowing either way.
2. **Does any `ExternalPrincipalExchange` caller set `client_id` in `additional_claims` today?** Reserving it turns a silently-honoured claim into a silently-dropped one. Nothing in this repo does; deployment config isn't visible to me.

### Verification

```
go vet          clean
golangci-lint   0 issues (v2.13.2 — the version CI runs; I'd only run go vet before, which is how the last CI failure surprised me)
go test ./...   9 packages ok, 0 FAIL (incl. tests/integration)
```

Re-run in full *after* the rebase, not just before it.

PR #333, opened 2026-09-07
fix: updated identities.list() to user .iter_all() because pagination support is for API not for sdk surface

KunalJavelin:
## Summary

---

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (does not change functionality, e.g. code style improvements, linting)
- [ ] This change requires a documentation update

---

## Testing
- [x] Manually tested
- [ ] Unit tests added/updated
- [ ] No tests required

---

## Impact / Risks


examples/a2a_org_hierarchy/provision.py
```diff
@@ -79,23 +79,14 @@ def _build_actor_token(wimse_uri: str, private_key_pem: bytes, aud: str) -> str:
...
-    identities.list() uses the server's default limit (20) and orders by
+    identities.list() returns a single page (limit 20 by default) ordered by
     created_at DESC, so the first-registered identities fall off the end.
-    We page manually via the transport to get the full set.
+    iter_all() walks every page at the service ceiling of 100.
     """
-    result: dict[str, tuple[str, str]] = {}
-    offset, limit = 0, 100
-    while True:
-        resp = client._transport.request(
```
KunalJavelin, line 89: @jalbrethsen-highflame Let's not use a private SDK function to call the server. We should use the exposed SDK client to call the underlying API with the right parameters. No calls should be made directly to the server via the client. If a client is doing this, it means we haven't exposed the API via the SDK.

saucam (approved):

examples/a2a_org_hierarchy/provision.py
rsharath: Agree

PR #347, opened 2026-09-11
feat: implement RFC 7523 §2.2 private_key_jwt client authentication (#206)

### LINKED ISSUE #206: private_key_jwt client authentication: enforce what registration already accepts (tls_client_auth deferred)
## Context

Raised by external review on #198 (https://github.com/highflame-ai/zeroid/pull/198#discussion_r3391429915): the token/introspection/revocation endpoints support only `client_secret_post` + `client_secret_basic` (the latter added in #198), while `private_key_jwt` and `tls_client_auth` exist as standard methods.

## The actual problem: private_key_jwt is half-implemented

Client registration **already accepts** `token_endpoint_auth_method: private_key_jwt` and stores key material (`jwks` / `jwks_uri` — see `CreateOAuthClientInput`, internal/handler/oauth_client.go; the service stores the caller's auth method verbatim with no allow-list), but:

- No `client_assertion` / `client_assertion_type` (RFC 7523 §2.2) validation exists anywhere — a client registered for key-based auth cannot actually authenticate with its key.
- **Silent downgrade — empirically verified** against the #198 branch. A client registered with `confidential: false, token_endpoint_auth_method: private_key_jwt, jwks: {...}` is stored as `client_type=public` with no secret, and `TokenEndpointAuthMethod` is never consulted at auth time:
  - `authorization_code` exchange with **no client authentication at all** → **200, token issued** (`verifyConfidentialClientAuth` treats no-stored-secret as public-client-OK; same path gates refresh_token and CIBA redemption).
  - Introspect/revoke with `client_id` only, no secret, no assertion → **accepted** as valid public-client auth (`VerifyPresentedClientAuth`).
  - `client_credentials` with no secret → 401 — this grant happens to **fail closed** (bcrypt compare against the empty stored hash always errors), so no token-issuance bypass there. The downgrade is real but does not extend to client_credentials.
- Doc surface is inconsistent: oauth_client.go advertises `private_key_jwt` in its enum; DCR (dynamic_registration.go) constrains to `client_secret_post`/`client_secret_basic`.

## Proposed scope

1. Accept `client_assertion` + `client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer` on /oauth2/token (and introspect/revoke), validated against the client's registered `jwks`/`jwks_uri` — RFC 7523 §3 claims (iss=sub=client_id, aud=issuer, exp, jti single-use). The `jwt_bearer` *grant* (§2.1) already implements the assertion-validation machinery; this is the §2.2 client-auth flavor of the same thing.
2. Enforce the registered `token_endpoint_auth_method`: a `private_key_jwt` client presenting a secret (or nothing) is rejected — closes the authcode/refresh/CIBA/introspect/revoke downgrade above. Until (1) lands, an interim hardening could simply reject auth-time use of clients whose registered method has no implementation.
3. Validate `token_endpoint_auth_method` at registration against an allow-list (today any string is stored verbatim).
4. Advertise `private_key_jwt` in `token_endpoint_auth_methods_supported` / `introspection_…` / `revocation_…` only once enforced. (Note: metadata currently also advertises `client_secret_basic` for the **token** endpoint, which has no Basic parsing — same fix family, can ride along.)
5. Update DCR to optionally allow `private_key_jwt` with `jwks`/`jwks_uri` in the registration payload.

## Why this is worth doing (and tls_client_auth is not, yet)

`private_key_jwt` fits ZeroID's no-shared-secrets-for-agents posture, reuses existing RFC 7523 machinery, and fixes an advertised-but-unenforced surface. It's also the direction OAuth 2.1 / FAPI point.

**`tls_client_auth` (RFC 8705) is explicitly deferred**: it requires an mTLS termination story (ZeroID deployments typically sit behind TLS-terminating proxies/ingress), certificate-bound access tokens, and deployment documentation — a feature program, not a gap fix, with no demand signal yet. Revisit when X.509 SVID / WIMSE mTLS work creates a concrete need. `client_secret_jwt` is skipped permanently (dropped in OAuth 2.1).

saucam:
Closes #206. Finishes what #346 started: that change made a client registered for `private_key_jwt` fail *loudly* instead of authenticating with nothing; this one makes it actually authenticate, and closes the registration-side hole that let the unenforceable state exist.

## All five scope items

| # | What landed |
|---|---|
| 1 | `client_assertion` / `client_assertion_type` accepted on `/oauth2/token`, `/introspect`, `/revoke`, verified against the client's `jwks` / `jwks_uri`. Full RFC 7523 §3 claim set; `jti` single-use against the shared replay table (namespaced `cla:`). |
| 2 | The registered `token_endpoint_auth_method` is binding. A key-based client cannot authenticate with a secret or with nothing. |
| 3 | `token_endpoint_auth_method` validated at registration against an allow-list. It was previously stored verbatim — any string at all. |
| 4 | Advertised in all three `*_endpoint_auth_methods_supported` lists, and only now that it is enforced. |
| 5 | DCR accepts `private_key_jwt` with `jwks`/`jwks_uri`, and issues **no** `client_secret` for such a client. |

Item 4 carries the rider the issue flags: the token endpoint now actually *parses* `client_secret_basic`. It had been advertised there with no Basic parsing at all, so a client following our own discovery document had its `Authorization` header silently ignored. That failed closed, but made an advertised method unusable.

## Twelve defects found during the work

Across four review rounds (security, cross-repo, an independent review, and an adversarial pass). **None came from reading the diff as written** — every one needed either a test, a query against live data, or an attack.

1. **The feature reopened the downgrade #346 closed.** Adding `private_key_jwt` to the implemented-methods set meant the old guard stopped firing on the introspection no-secret branch — and such a client lands `client_type=public` with an empty secret hash, so it reached that branch and would have authenticated with its `client_id` alone. Caught by the integration test.

2. **A pre-existing hole in #346.** The CIBA `bc-authorize` and token-poll paths never called `rejectUnimplementedClientAuth` at all, so a `tls_client_auth` client could *already* initiate CIBA unauthenticated. `bc-authorize` fires the deployer's notifier, so that means real approval prompts at arbitrary users under that client's identity.

3. **A `private_key_jwt` client could hold — and use — a `client_secret`.** `confidential` and `token_endpoint_auth_method` are independent inputs and the secret is minted off the former before the latter is read. The resulting client was refused on the grants that enforce the method and **accepted** on `client_credentials` and ID-JAG, which verified the secret directly. The weaker credential won wherever nothing checked. A comment in the first commit asserted this couldn't happen; it was wrong. Now unrepresentable at the root, with the two grants routed through shared enforcement as defense-in-depth for older rows.

4. **A `jwks_uri`-only client could never authenticate.** `jwks` is nullable `jsonb` and bun writes a nil `json.RawMessage` as JSON `null`, so such a client reads back with four bytes of `"null"` and looked like it carried *both* key sources. Found by querying dev1/prod, not by reading. No in-memory unit test could catch it — a hand-built `OAuthClient` has a genuinely nil JWKS; a database round trip does not. The regression test persists and reloads.

5. **An interop bug that would have shipped silently.** jwx requires a `kid` header when verifying against a key set, but `kid` is optional per RFC 7515 §4.1.4 and clients publishing one key routinely omit it. `private_key_jwt` would have worked only for clients that happen to set one — the same shape as #323. The integration test now deliberately mints kid-less assertions.

6. **`jwks_uri` accepted `http://`.** The key set at that URL decides who may authenticate as the client, so over cleartext anyone on the path can substitute it and mint assertions at will.

7. **`ClientJWKSCache.get` held the cache mutex across a synchronous fetch** of a client-supplied URL — an unauthenticated way to stall every `private_key_jwt` verification server-wide.

8. **The jti ledger was keyed globally, not per client.** RFC 7519 §4.1.7 makes jti uniqueness the *issuer's* responsibility, so two clients may legitimately pick the same value — and one locked out the other with an error reading as an attack on the victim. No adversary required: timestamp-derived jti collides in the same second.

9. **An over-long jti returned HTTP 500** — `VARCHAR(512)`, SQLSTATE 22001, not the duplicate-key sentinel. Caller-controlled input producing a 5xx on the public token endpoint.

10. **One RFC 7592 PUT silently and permanently bricked a key-based DCR client**, and `GET` never returned `jwks`/`jwks_uri`, so the client could not read what a full-replacement PUT required it to restate.

11. **The replay row expired before the assertion did.** `WithAcceptableSkew` keeps an assertion verifiable 60s past `exp`, but the row was written with the bare `exp` — and `dpop.MemoryStore` treats an expired entry as absent and overwrites it, making the second redemption succeed *deterministically*.

12. **Advertising `private_key_jwt` for introspection/revocation broke RFC 8414 §2**, which makes the matching `*_auth_signing_alg_values_supported` members REQUIRED once it appears. Self-defeating for a change whose point is that the advertisement is honest.

Two claims in my own commit messages were also false and are corrected: *"namespaced so they can never collide"* (DPoP inserted a raw client-chosen key into the same table) and *"unrepresentable at the root"* (the guard sat at two of three callers of that write).

## Blast radius — measured, not assumed

`SELECT` over dev1 and prod: **every** client is registered `token_endpoint_auth_method=none`, which the new allow-list accepts unchanged. A cross-repo sweep of all sibling repos found no registration site that passes anything else — every caller either omits the field or sends `"none"`.

## Behaviour changes third-party clients may notice

- **`Authorization: Basic` + body `client_id` on `/oauth2/token` is now a 400.** Previously the header was ignored entirely and the body credentials won. RFC 6749 §2.3 forbids using two authentication methods in one request. No in-workspace caller does this; a third-party client that sends both will change behaviour.
- Registration rejects `token_endpoint_auth_method` values outside the allow-list with 400, where any string was previously stored verbatim. Measured as affecting nothing deployed (see above).

## Verification

`go build`, `go vet`, `gofmt` and `golangci-lint` (v2.13.2, the version CI pins) all clean. Full suite green including the ~52s integration run against testcontainers Postgres. 27 new unit tests, 7 new integration tests.

## Deliberately not in this PR

- **CIBA client-assertion support.** Those paths carry no assertion to verify, so a key-based client is *refused* there rather than silently waved through. Accepting one (CIBA Core §7.1 permits it) is follow-up.
- **`highflame-authn` config plumbing.** It hand-builds `zeroid.Config` as a struct literal, so `client_auth` is unreachable until mapped there. Not a break — the zero value is the safe production default — but it means there is no way to enable a localhost `jwks_uri` for local dev. Should land with the `go.mod` bump.
- **SDK support for `jwks`/`jwks_uri`.** No SDK exposes those fields, so a caller cannot satisfy a `private_key_jwt` registration. Fails closed; capability gap.
- **Pre-existing CIBA validations in `RegisterClient`** still return 500 for caller-metadata errors. Same bug class as the 400 mapping fixed here; left alone to keep this diff honest.


saucam:
<details>
<summary>Auto-fixed by pr-shepherd (iteration 1): 1 lint finding</summary>

**`highflame-lint-check`** failed on a single `staticcheck QF1001` finding in `internal/service/oauth_client.go:462` (the `jwks_uri` scheme validation). Rewritten via De Morgan's law:

```go
-if u.Host == "" || (u.Scheme != "https" && !(allowPrivateJWKS && u.Scheme == "http")) {
+if u.Host == "" || (u.Scheme != "https" && (!allowPrivateJWKS || u.Scheme != "http")) {
```

`!(A && B)` becomes `!A || !B`, so this is logically equivalent and there is no behaviour change. `jwks_uri` still must be an absolute `https://` URL, with `http://` accepted only when the private-JWKS escape hatch is enabled. No other file touched.

Verified locally before push, in a worktree at this commit:

- `golangci-lint run ./...` at **v2.13.2** (the version CI pins): `0 issues`
- `go build ./...`, `go vet ./...`, `gofmt -l .`: clean
- `go test -count=1 ./...`: all 9 packages pass, including the 52.7s `tests/integration` suite
- `go test -race` on `internal/service` and `internal/handler`: clean

Branch was already up to date with `main` (0 behind), so no rebase was needed.

Re-running CI. This is a mechanical change with no behavior impact.
</details>

rsharath (approved):

saucam:
Follow-up tracked in #348 — `client_type` no longer implies "holds no credential", and this PR defends that by enumeration (four separate call-site patches, two added here) rather than by construction. Deliberately not folded in: it is a type-model change, nothing is exploitable today, and it would have doubled the size of an already-large diff.

PR #354, opened 2026-09-15
fix: make zeroid resolvable from the proxy — real tags for both submodule pins

saucam:
**zeroid cannot currently be consumed.** In a fresh module:

```
$ go get github.com/highflame-ai/zeroid@v1.9.3   # and v1.7.1, and every release between
go: github.com/highflame-ai/zeroid imports
    github.com/highflame-ai/zeroid/pkg/authjwt: reading .../go.mod
    at revision pkg/authjwt/v0.0.0: unknown revision
```

In-repo builds never notice, because the local `replace` hides it — and our only consumer (highflame-authn) already pins `pkg/authjwt` explicitly, so nobody hit it.

## Two pins, one root cause

Consumers **ignore `replace` directives in dependency modules**, so the `require` lines are exactly what the proxy serves.

| pin | before | after | why |
|---|---|---|---|
| `pkg/authjwt` | `v0.0.0` | `v1.9.2` | placeholder — unresolvable for consumers |
| `pkg/dpop` | `v1.6.2` | `v1.6.3` | real tag, but stale: v1.6.3 carries the replay-store namespacing fix from #347 |

Neither submodule's source has drifted from the tag now referenced, so this is a pure pin correction — no release dance needed.

## Why the `authjwt` placeholder was there, and why it stopped being safe

RELEASING.md documented an asymmetry: `pkg/dpop` is imported by non-test code so it needs a real tag, while `pkg/authjwt` was imported *"only from `tests/integration/`"* and Go doesn't follow test imports across module boundaries — so its pin was *"invisible to downstream consumers and never needs to be a real version."*

**That was true when written.** It stopped being true on **2026-06-19**, when #211 (direct OIDC IdP federation) added `internal/service/external_issuer_registry.go` and the `server.go` option plumbing — both importing pkg/authjwt from non-test code. `internal/service/client_jwks.go` joined later with `private_key_jwt` (#347), adding an importer to an already-broken state.

Nothing fails at the moment such an import is added; it surfaces only when an outside consumer tries to resolve the module.

**So the blast radius is every release from v1.7.1 onward — roughly three months — not just v1.9.3.** Verified directly rather than inferred:

```
$ go get github.com/highflame-ai/zeroid@v1.7.1
go: ... reading .../pkg/authjwt/go.mod at revision pkg/authjwt/v0.0.0:
    unknown revision pkg/authjwt/v0.0.0
```

That nobody noticed for three months is itself the argument for the guard: nothing in the repo could surface it, and our only consumer (highflame-authn) pins `pkg/authjwt` explicitly.

## The drift guard worked, and still couldn't save us

Worth recording, because it changes what the guard is *for*.

The **v1.9.3 release run failed** at exactly `Verify pkg/dpop/ source matches the version in go.mod` — correctly, since dpop source had drifted from the pinned v1.6.2. But **for a Go module, pushing the tag *is* publication**; the workflow only runs afterwards.

So v1.9.3 went out as a **half-release**: the module tag is live and serving a broken `go.mod`, while `tag-submodules`, `goreleaser` and the Docker build were all skipped. v1.9.3 has **0 release assets** against v1.9.2's **5** — which is why `pkg/authjwt/v1.9.3` doesn't exist.

## Changes

- **`go.mod`** — both pins point at real published tags, with the history recorded where the next person will be editing.
- **`release.yml`** — the guard now covers **both** submodules instead of `pkg/dpop` alone, and rejects a `v0.0.0` placeholder outright with the consumer-facing symptom in the error text. Its comment no longer claims `pkg/authjwt` is test-only.
- **`RELEASING.md`** — the asymmetry section is *corrected rather than deleted*, so the reasoning that went stale stays visible. The "test-only nested module" recipe no longer recommends a `v0.0.0` placeholder, since a module's test-only status is not a property anyone maintains.

### Guard verified by mutation

Ran the guard logic against three trees:

| state | result |
|---|---|
| fixed | passes |
| `authjwt` reverted to `v0.0.0` | **fails** — placeholder rejected |
| `dpop` reverted to stale `v1.6.2` | **fails** on drift — i.e. it reproduces the original bug |

## Follow-up this does *not* do

**v1.9.3 remains half-published.** A **v1.9.4** cut after this merges gets a clean release with both submodule tags and artifacts.

## Verification

`go build`, `go vet`, `gofmt` clean; full unit + domain suites green under `-race`.


akashjavelin (approved):

akhiljavelin (approved):

saucam:
Went through all nine Oracle comments. **Seven are resolved by the rework Oracle could not see** — its review is against head `87283d0`-minus, before the lockstep redesign. Two were live and are fixed in `c01c2a9`, along with a third Oracle did not raise.

## Answering the question five comments hinge on

Oracle asked repeatedly whether `pkg/jwks` has its own `go.mod`, and correctly said the answer decides whether the PR fixes the bug or reintroduces it. **It does have one — and it is handled:**

| | |
|---|---|
| `pkg/jwks` pinned in root `go.mod` | ✅ `v1.9.4`, not a placeholder |
| Covered by the release guard | ✅ `for MOD in dpop authjwt jwks` |
| Tagged at release | ✅ `tag-submodules` covers all three |
| Is it a true leaf? | ✅ its `go.mod` requires nothing from zeroid |

That last one resolves the **module cycle** flagged on `jwks_alias.go:7`. `pkg/authjwt` requires `pkg/jwks`, and `pkg/jwks` requires nothing back — which is exactly the "cleaner endgame: give pkg/jwks its own go.mod as a leaf submodule" Oracle proposed. Root → {jwks, dpop, authjwt}, authjwt → jwks, no cycle. No tracking issue needed; it is already the shape.

The `v0.0.0`-placeholder concern on `go.work:26` is likewise moot — no pin is a placeholder, and the guard now rejects any pin that is not exactly the release version, which is strictly stronger than rejecting `v0.0.0`.

## The drift concern was real and is now structurally impossible

`jwks_alias.go:4` was right that the file drifted `pkg/authjwt` from `v1.9.2` — I confirmed it (`jwks.go` deleted, 325 lines; `jwks_alias.go` added). Under the old pin-at-last-release scheme that would have failed the guard.

Lockstep removes the failure mode rather than detecting it: pins name the version **being released**, tagged at that same commit, so "go.mod references a stale submodule" has no representable state. The guard checks pins equal `RELEASE_TAG` **before** any tag is created.

This also dissolves a deadlock I raised in my earlier audit of this PR — under the previous design, the first edit to `pkg/authjwt/` would have blocked releases with no way to satisfy the guard, since `pkg/authjwt` tags are only created during a release. Lockstep makes that unreachable. Worth recording that the redesign fixed a defect beyond the one it targeted.

## What was still broken — fixed in `c01c2a9`

**1. Tag push order was dependency-inverted.** `tag-submodules` pushed `authjwt, dpop, jwks`, but `pkg/authjwt@vX.Y.Z` requires `pkg/jwks@vX.Y.Z` — so authjwt was published before the dependency it names existed. Normally milliseconds. But pushing a tag *is* publication, which is this PR's own central lesson: if the job dies between those pushes, `pkg/authjwt` is permanently published unresolvable — the same failure one level down, found the same way, by an outside consumer. Now leaf-first: `jwks, dpop, authjwt`.

**2. `go.mod` still opened with the comment this PR exists to disprove.**

```
// pkg/authjwt is import-only-from-tests. Go does not follow test
// imports across module boundaries, so the v0.0.0 placeholder in the
// require directive above is invisible to downstream consumers.
```

Every clause false: `pkg/authjwt` is imported from non-test code (the thesis), and there is no `v0.0.0` placeholder — the require three lines above says `v1.9.4`. It sat directly on top of the LOCKSTEP paragraph contradicting it. This is not an ordinary stale comment; it is *the* stale comment, the one whose going-unnoticed made every release from v1.7.1 unresolvable, and the text RELEASING.md now quotes as a cautionary tale. Deleted. Also added `pkg/jwks` to that paragraph's module list, which named only authjwt and dpop while the pins above it and the guard both cover three.

**3. Typo Oracle did not raise:** `one forward jump (pkg/dpop v1.9.4 -> v1.9.4)` — a find/replace artifact. The jump is from `v1.6.3`; as written the sentence argues for a cost it never names.

## Left for you

- **`jwks_alias.go:22`** — mechanically diffing the pre-move exported surface for error sentinels and interfaces. Worth doing; I did not, and the compiler will not catch a missing `errors.Is` sentinel until cerberus/shield/authn next bump.
- **`external_issuer_registry.go:62`** — tests for the three contract properties (best-effort warm-up with deferred fail-closed, last-wins `WithHTTPClient`, idempotent `Close`). If `pkg/jwks` is a verbatim move these are inherited; if reworked, they are worth pinning, and the SSRF-guard injection is the reason.
- **`jwks_alias.go:41`** — the extra SSRF line on `WithHTTPClient`. Cheap and I would take it.
- **`pkg/dpop` v1.6.3 → v1.9.4** is a version discontinuity for its direct consumers (Cerberus, Shield, Firehog pin it independently). Deliberate per the design, but it is a coordination item, not just a number.

Verified before push: `GOWORK=off` build clean — the path that masks pin problems, so the one worth running — plus `go vet`, `gofmt`, and `release.yml` still parsing as YAML.

